PR #11, opened 2026-01-29
Add more clubs to the club picker

trotzig:
I was playing around with the UI and noticed that the club selection was a little bit limited. While it's probably okay to use e.g. 5-wood when practicing with a hybrid, I thought it would be better to have the right club in the list. The list is of course not complete ("where's my 1-iron?") but it should be enough for 99.5% of golfers.

I picked enough clubs so that they would be divisible by four. The grid layout looks best with these groups.


jewbetcha (approved):
LGTM - thanks for the contribution @trotzig !

PR #13, opened 2026-01-30
Add visual tests through Happo

trotzig:
While the surface area of the user interface isn't super big, having a visual log and a visual quality check makes development easier. By rendering the App component in different variants we also make sure that we don't accidentally break anything.
    
The Happo test suite is set up to use the Storybook as a test suite. Anything added to Storybook will be subject to UI testing.
    
I've added a single chrome target for now with the viewport set to a 7 inch Raspberry PI kiosk screen such as the one listed in the Raspberry Pi Setup guide. If we think that people will use the UI in other viewports and platforms we can add more targets to cover that.

### DIFF
--- a/.github/workflows/ui-build.yml
+++ b/.github/workflows/ui-build.yml
@@ -40,3 +40,32 @@ jobs:
       - name: Lint
         run: npm run lint
         working-directory: ui
+
+  happo:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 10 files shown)


jewbetcha:
> Nope -- I've set up a free open source account for Openflight at Happo.

Great! I'll grab the required keys later today

trotzig:
Awesome -- I added you to the Happo account. If you go here you should be able to grab the API keys: https://happo.io/a/1566/account -- let me know if that doesn't work for some reason.

PR #22, opened 2026-02-26
Resample irregular spin detections before FFT

aadnesd:
## Summary
- make rolling-buffer spin detection timestamp-aware
- resample irregularly spaced post-impact speed detections onto a uniform grid before FFT
- pass per-reading timestamps from `process_capture()` into `detect_spin()`
- add a regression test showing improved RPM accuracy with dropped detections

## Why
The spin FFT previously treated filtered detections as uniformly sampled even when detections were dropped, which can bias RPM estimates and create false peaks.

## Testing
- `PYTHONPATH=src /Users/adne.skjelbreid.djuve/Documents/Codeprojects/openflight-pr-spin-confidence/.venv/bin/pytest tests/test_rolling_buffer.py -v -k "detect_spin_resamples_irregular_timestamps_before_fft"`


jewbetcha:
thank you @aadnesd ! will test this out asap

PR #26, opened 2026-03-01
chore: add prettier configuration for consistent UI formatting

### LINKED ISSUE #25: Add prettier config to ui for consistent formatting
# Prettier config for UI

> See: [Prettier](https://prettier.io/)

## Description

The UI has mixed formatting throughout. My suggestion is to add a standard-ish `.prettierrc` config so that current and future collaborators can contribute while maintaining a consistent standard format. 

This is my suggested config:

```json
{
  "printWidth": 120,
  "trailingComma": "es5",
  "tabWidth": 2,
  "semi": true,
  "singleQuote": true
}
```

This is mostly based on the fact that most of the files in `/ui` are single-quoted and use trailing semicolons... but not all of them! Hence, the need for a configuration file.

robotnoises:
## Description

Adds `prettier` config and runs a global format on the full `ui` for consistent formatting on UI contributions. This is a somewhat opinionated config based on what I observed on the `ui`.

See [the prettier docs for more info](https://prettier.io/docs/)

## Summary of changes

- Adds `.prettierrc` and `.prettierignore` files to the root of `ui`
- Ran `npx prettier . --write` from the root of `ui` which will reformat all files to match the config
- Small update to the readme file

## Screenshots

> Didn't break it!!

https://github.com/user-attachments/assets/1e97db5b-802a-4fc8-a70e-3db8765e4da4

## Notes

- Most of the changes are just formatting stuff, there were no functional changes introduced in this PR
- Closes #25

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,6 +411,8 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for full guidelines. Quick summary:
 - **Tests required**: Run `pytest tests/ -v` before submitting a PR
 - **Code quality**: Run `pylint src/openflight/` (score 9+)
 - **UI changes**: Run `cd ui && npm run build` to verify
+- **UI formatting**: If your IDE does not include automatic "format-on-save," run `cd ui && npm run format` to format TypeScript/React files. 
+  - Note: editors like Zed, WebStorm, and VS Code (with the [Prettier extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)) will pick it up automatically and format when saving.
 
 ## Documentation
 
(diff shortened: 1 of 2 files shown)


ui/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
```
robotnoises: If you don't agree with these settings I'm happy to change and re-format. This was my best guess based on what already existed in the repo.

jewbetcha (approved):
LGTM! Thanks for this. There may be a time later down the line where we tune the rules but I think it's great for now

PR #33, opened 2026-04-02
Adding Compact 3D Printable Case

InvertedAviatorsRC:
Adding all files, project, and readme necessary for 3d printed case build. If you could add the screws necessary to mount all the sensors/screen/pi board to the 3d printed mounts that would be great!


jewbetcha:
Thanks for this! Could you please add them under the `cad/` directory, and then create a new directory for this build in particular?

PR #37, opened 2026-04-03
feat(kld7): noise filtering and per-club dynamic angle gates

brandonh94:
## Summary

- **Swing precursor filter**: far-range detections with no preceding close-range swing activity are rejected as noise
- **Shot timestamp window**: only consider ball-range frames within +-50ms/+200ms of OPS243 shot detection, eliminating post-net scatter and pre-shot blips
- **Per-club dynamic angle gates**: replaces the fixed 40 degree ceiling with per-club (min, max) windows based on TrackMan launch data (e.g. driver: 4-22 deg, 7-iron: 13-34 deg, LW: 26-52 deg). Adds a floor so sub-threshold readings (e.g. 3 deg on a 7-iron) are rejected as noise
- **Horizontal orientation bypass**: vertical gates are skipped for horizontal-mount configurations where negative angles are valid
- **Analysis scripts**: compare_kld7_filters.py, show_swing_times.py, show_accepted_swings.py for offline pkl capture analysis

## Test plan

- [ ] uv run pytest tests/test_kld7.py -v -- all 29 tests pass, 2 skipped (hardware)
- [ ] Run scripts/compare_kld7_filters.py against pkl captures to verify old vs new detection quality
- [ ] Live test with K-LD7 hardware hitting 7-irons and wedges

Generated with Claude Code


jewbetcha:
Closing this for now, can re-open later if need be

PR #45, opened 2026-04-10
Add metric unit toggle to UI

### LINKED ISSUE #17: Add option for Meters/kmh
UI should have an option to switch to metric

PhilipCaliss:
Fixes #17


jewbetcha (changes requested):
Thanks for the contribution! The Context/Provider/hook pattern is clean and matches the existing `ShotProvider` pattern well. localStorage persistence is a nice touch. A few things to address:

## 1. `°` → `deg` and `—` → `--` are unrelated visual regressions

The PR changes all degree symbols from `°` to `deg` and em-dashes `—` to `--` for missing values throughout `ShotDisplay.tsx` and `ShotList.tsx`. These aren't related to the unit toggle — angles are the same in both unit systems. This makes the display wider/noisier for all users. Please revert these back to `°` and `—`.

## 2. StatsView stat cards need unit labels

The stat cards ("Avg Ball", "Max Ball", "Avg Carry", "Avg Club") don't display units anywhere. Before this PR it was implicitly mph/yds, but now that users can switch systems, a user in metric mode sees bare numbers with no indication whether it's km/h or meters. Please add the unit to each stat card.

## 3. Use the `useUnitPreference` hook directly instead of prop drilling

`unitSystem` is passed as a prop through `App → ShotDisplay/StatsView/ShotList`, but there's already a context hook (`useUnitPreference`) set up for exactly this. The existing pattern in this codebase (e.g., `useShotContext`) is to consume context at the point of use. Each component should call `useUnitPreference()` directly rather than accepting `unitSystem` as a prop — this removes the coupling through `App.tsx` and means you won't need to touch the parent every time a new child needs unit info.

jewbetcha:
Thank you! Last thing - could you add some tests? Would be great. Then will merge!

jewbetcha:
thank you @PhilipCaliss ! will merge shortly. 

Real quick, can I get some written approval from you on the license change from MIT to AGPL? Just happened the other day

jewbetcha (approved):

PhilipCaliss:
> thank you @PhilipCaliss ! will merge shortly.
> 
> Real quick, can I get some written approval from you on the license change from MIT to AGPL? Just happened the other day

Sounds good.

PR #61, opened 2026-04-13
Feat/ballistics physics model

JedS:
## Summary

Adds a physics-based ball flight simulator and the tooling needed to validate it against TrackMan ground truth.

* **`src/openflight/ballistics.py`** — RK4 integrator (500 Hz) with gravity, drag, and Magnus lift. Coefficient fits from Bearman & Harvey (1976) / Kensrud & Smith (2018) wind-tunnel data: `Cd(Sp) = 0.205 + 0.18·Sp` and `Cl(Sp) = 0.32·Sp/(0.15+Sp)`. Spin decays at 4%/s.
* **`resolve_launch()`** — deterministic spin-fallback policy: uses measured spin when `spin_confidence ≥ 0.7`, otherwise substitutes club-typical RPM (TrackMan PGA Tour values). Tags output via `spin_source: "measured" | "club_typical"`.
* **`LaunchConditions` / `Trajectory` dataclasses** — clean handoff shape for downstream consumers (external simulator or UI).
* **`server.py`** — when the simulator is enabled, carry is computed from it whenever a vertical launch angle is available; otherwise the old table-based `estimate_carry_with_spin()` runs (also when angle is missing).
* **`--ballistics` server flag** _(plus the matching toggle in `start-kiosk.sh`)_ — opt-in switch that enables the physics simulator for live shots. Default is **off**: until coefficients are validated against more TM data, all non-mock shots go through the legacy table estimator. The chosen mode is announced at startup and the table-fallback log line identifies which condition triggered it (`no launch angle` vs `ballistics disabled`).
* **`tests/test_ballistics.py`** — 18 tests covering launch resolution, carry/apex ranges per club, launch-angle and spin sensitivity, fade/draw/straight spin-axis behavior, spin decay, landing angle, and edge cases.
* **`scripts/analysis/validate_ballistics.py`** _(new)_ — runs the model on TrackMan inputs (physics-only check) and on OpenFlight inputs (end-to-end check), compares each prediction to TrackMan flat carry, emits per-shot CSVs, per-club + per-session stats tables, and scatter plots.
* **`scripts/analysis/sweep_ballistic_coeffs.py`** _(new)_ — differential-evolution + Nelder-Mead fit of the four aero constants against TrackMan carry; `--loso` flag adds leave-one-session-out cross-validation so we can see whether fitted constants generalize.
* Both analysis scripts accept multiple `--trackman` / `--comparison` files with matching `--session-label` values so additional sessions can be folded in without code changes.

## Motivation

Current carry estimation in `launch_monitor.py` is a club-typical lookup table. This module gives OpenFlight a real aerodynamic model suitable for (a) accurate standalone carry/apex/trajectory display and (b) feeding an external simulator with `{ball_speed, launch_angle_v, launch_angle_h, spin_rpm, spin_axis}`.

Spin detection on the OPS243-A is only ~50–70% reliable, so the fallback policy is essential: rather than propagating noisy spin into the physics model, low-confidence shots fall back to club-typical values, keeping trajectory output stable.

Once the model is live we need a way to **prove** it's right. Running this PR against a paired OpenFlight + TrackMan session from 2026-05-06 (24 shots: driver / 7-iron / PW) exposed a sizeable, systematic per-club bias on the default coefficients:

| Club  | Default model bias | Default RMSE |
|-------|--------------------|--------------|
| Driver | −37.2 yd          | 38.6 yd      |
| 7-iron | +11.0 yd          | 11.6 yd      |
| PW     | +11.0 yd          | 13.6 yd      |

i.e. with TrackMan's own measurements as input, the model still undershoots a driver by an average of 37 yards. The validation + sweep scripts make that visible and tunable, but tuning is deferred to a follow-up PR — see Design notes below.

## Design notes

* **Pure physics, no ML.** No training data available; textbook coefficients with published citations.
* **Deterministic output.** No probabilistic ranges — the downstream simulator/UI contract is a single trajectory per shot.
* **Default constants kept as published.** A 4-parameter sweep on the 24-shot 2026-05-06 session cuts overall RMSE from 24.5 yd → ~4 yd, but that is 4 free parameters against 24 data points across only 3 clubs — high overfit risk. The sweep script and its `--loso` flag are checked in so the tuning can be done honestly once at least one more OF+TM session is recorded; until then we keep the published Bearman-Harvey / Kensrud-Smith forms and only ship the validation tooling.
* **Opt-in until validated.** The simulator ships disabled by default because the 24-shot driver bias above is large enough that turning it on globally would regress what users see on screen. `--ballistics` enables it explicitly for A/B work and for users running tuned coefficients locally; the table estimator remains the production path until more OF+TM sessions are recorded and the sweep is rerun. The chosen mode is logged at startup and the table-fallback log line identifies which condition triggered it.
* **Air-density normalization.** TrackMan's "Flat" carry assumes 77 °F / 0 ft / no wind. The validation script passes `air_density = 1.184 kg/m³` to `simulate()` so the comparison is apples-to-apples; `simulate()`'s default stays at the 15 °C ISA value of 1.225 kg/m³ for normal use.
* **Multi-session ready.** Both analysis scripts accept N `--trackman` files (and N matching `--comparison` files). Each shot is tagged with a session label so the per-shot CSVs, stats blocks, and scatter plots stay legible as more data lands.
* **No patent/licensing concerns.** Coefficient forms and values are public-domain wind-tunnel literature.

## Test plan

* CI pytest passes (`pytest.yml`)
* CI pylint ≥ 9.0 (`pylint.yml`)
* Driver carry falls in 250–300 yd range for typical launch conditions
* 7-iron carry falls in 140–200 yd range
* Positive spin axis produces right lateral, negative produces left, zero stays straight
* `resolve_launch` returns `spin_source="measured"` only when `spin_confidence ≥ 0.7`
* `validate_ballistics.py` against `session_logs/OpenFlight-Test.Normalized.csv` reproduces the bias table above and emits TM-inputs / OF-inputs scatter PNGs with no errors
* `sweep_ballistic_coeffs.py` converges (DE + NM) on the same input and reports a 24.5 → ~4 yd RMSE improvement with the optimal coefficients printed for later review
* Both scripts accept a multi-session invocation (`--trackman A B --session-label may06 jul10 [--loso]`) and produce per-session stats / markers without code changes
* Without `--ballistics`, `resolve_launch` / `simulate` never run for live shots — pinned by `test_carry_skips_ballistic_when_ballistics_disabled` (default) and `test_carry_uses_ballistic_simulator_when_launch_angle_present` (opt-in) in `tests/test_server.py`


jewbetcha:
Thais is sick! Thanks so much. Will test first thing tomorrow

jewbetcha:
@JedS just to cover our bases, we recently changed the license to AGPL v3. Just want to get your written approval on this (since I think this PR was made with MIT still)

jewbetcha:
Nice work — physics model looks solid and the spin fallback policy is the right call for OPS243-A reliability.

One thing to address: the `0.7` spin confidence threshold now lives in three places (`launch_monitor.py:315`, `rolling_buffer/processor.py:598`, and the new `SPIN_CONFIDENCE_HIGH` in `ballistics.py`). `resolve_launch` also re-implements the confidence check rather than using `Shot.spin_quality == "high"`.

Suggest extracting a shared constant in `launch_monitor.py` and either importing it or using the existing `spin_quality` property in `resolve_launch`. Small change, prevents silent drift if the threshold ever moves.

JedS:
> @JedS just to cover our bases, we recently changed the license to AGPL v3. Just want to get your written approval on this (since I think this PR was made with MIT still)

New license is fine for me

JedS:
> Nice work — physics model looks solid and the spin fallback policy is the right call for OPS243-A reliability.
> 
> One thing to address: the `0.7` spin confidence threshold now lives in three places (`launch_monitor.py:315`, `rolling_buffer/processor.py:598`, and the new `SPIN_CONFIDENCE_HIGH` in `ballistics.py`). `resolve_launch` also re-implements the confidence check rather than using `Shot.spin_quality == "high"`.
> 
> Suggest extracting a shared constant in `launch_monitor.py` and either importing it or using the existing `spin_quality` property in `resolve_launch`. Small change, prevents silent drift if the threshold ever moves.

I can be wrong, but I think the 0.7 from processor.py is "the number stamped on medium-quality detections." It's a label, not a cutoff.
I shared SPIN_CONFIDENCE_HIGH between launch_monitor and ballistics.

However, the wire resolve_launch + simulate into the server shot pipeline is not yet done. Where carry_spin_adjusted is computed today (server.py:1096–1113), instead call resolve_launch(shot) → simulate(conditions). Store the Trajectory on Shot (new optional field, e.g. trajectory: Optional[Trajectory]). Fall back to the table estimator if resolve_launch returns  None (no vertical launch angle).

JedS:
Integration in server has been done : 
 Prefer the physics simulator (drag + Magnus, RK4) when a vertical launch angle is available; fall back to the table estimator only when the angle is missing (resolve_launch returns None).
No UI change.
TODO for exposing ballistics to UI : 
  - Add Shot.trajectory field (Optional[Trajectory]) and decimated point list to WebSocket payload.
  - UI component renders a 2D side-view trace + apex/landing markers.

JedS:
Looking at the https://github.com/jewbetcha/openflight/blob/main/session_logs/comparison_20260506_plots/compare_takeaways.png,
I wonder whether this ballistic model will help to refine the comparison of the carry.
As the model will take the spin effect into account, it will be interesting to compare the results once spin is measured correctly.

aadnesd:
planning to merge this one @jewbetcha ?

jewbetcha:
@aadnesd final thing to do here is to compare how this model stacks up with our "source of truth" which is trackman. Give me a couple hours!

JedS:
@jewbetcha I had an option to enable/disable the model from the **start-kiosk.sh**
And 2 new scripts:

- validate_ballistics.py that runs the model on TrackMan inputs (physics-only check) and on OpenFlight inputs (end-to-end check), compares each prediction to TrackMan flat carry, emits per-shot CSVs, per-club + per-session stats tables, and scatter plots.
- sweep_ballistic_coeffs.py that try to fit the model from the Trackman sessions.

The more comparison we can do, the more accurate the model will be.
It could be interesting to merge this (eventually, set the model off by default), so validation and sweep will be available for everyone.

jewbetcha:
@JedS perfect. One last request - instead of `--no-ballistics` could you change it to `--ballistics` instead? I would rather it be opt-in than opt-out just until we can test more. Then will merge 🚀  Thank you!

JedS:
> @JedS perfect. One last request - instead of `--no-ballistics` could you change it to `--ballistics` instead? I would rather it be opt-in than opt-out just until we can test more. Then will merge 🚀 Thank you!

done

jewbetcha (approved):

PR #66, opened 2026-04-22
fix(kld7): increase ring buffer to 6s

johnpacino:
The 2s default buffer (68 frames) was being overwritten during the ~4s OPS243 serial dump, causing missed KLD7 detections. Increasing to 6s (204 frames) gives enough headroom. Verified with 52+ shots at 100% detection rate.


jewbetcha (approved):
LGTM

PR #69, opened 2026-04-25
Fix rolling buffer shot timing

cduane-cx:
## What does this PR do?

This PR fixes the timing handoff between rolling-buffer shot detection and K-LD7 angle correlation. Before this change, the live path could correlate K-LD7 data against delayed callback time or scan too much of the ring buffer, which made shot matching less reliable. It also fixes rolling-buffer capture metadata so duration and timing stay correct when the radar is not running at the default sample rate.

### Before / after

- `impact_timestamp` on rolling-buffer shots
  - Before: a shot detected at ball impact could reach the server without an `impact_timestamp`, so K-LD7 correlation fell back to callback wall-clock time.
  - After: rolling-buffer shots carry a reconstructed impact timestamp based on capture timing and ball offset, so K-LD7 correlation is anchored to the captured impact moment.

- K-LD7 `shot_timestamp` handling
  - Before: `get_angle_for_shot(shot_timestamp=...)` accepted a timestamp but still scanned the full RADC ring buffer.
  - After: the tracker narrows RADC frame selection around the provided shot timestamp before extracting the angle.

- K-LD7 startup failure path
  - Before: if K-LD7 initialization failed, the server tried to call `sys.exit(1)` without importing `sys`, raising `NameError` instead of exiting cleanly.
  - After: the failure path exits cleanly with status `1`.

- Rolling-buffer capture duration metadata
  - Before: `IQCapture.duration_ms` always assumed `30 ksps`, so a `25 ksps` capture reported the wrong duration.
  - After: capture duration uses the stored sample rate from the actual capture, so non-default rates report correct timing.

## How was this tested?

- `uv run --python .venv\Scripts\python.exe --no-project pytest tests -q --basetemp=.pytest-tmp`
- `uv run --python .venv\Scripts\python.exe --no-project pylint src/openflight --fail-under=9`
- `npm.cmd run lint`
- `npm.cmd run build`
- `ruff` was run through pre-commit for the touched files and passed
- repo-wide `uv run ruff check src/openflight` still reports pre-existing unrelated issues in camera and package init modules

## Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -48,6 +48,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - K-LD7 launch-angle processing now uses OPS243 impact timestamps for live correlation
 - K-LD7 ball-burst selection now prefers coherent far-target paths instead of averaging all far PDAT detections
 - Live K-LD7 vertical launch angles now fall back to the existing club-and-speed estimate when the radar result is an obvious false positive
+- K-LD7 RADC extraction now narrows frame selection around the recorded shot timestamp instead of scanning the full buffer
+- Rolling-buffer capture metadata now preserves the configured sample rate and reconstructs impact timestamps from capture timing
 - Spin detection improved: Hann windowing, zero-padding to 256 points, band-limited search
 - All shot metrics (spin, launch angle, club speed, carry) always shown in UI
 - Shot logging unified — all metrics in single `shot_detected` entry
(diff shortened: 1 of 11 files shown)


jewbetcha:
Thanks for this. I will give it a test over the next couple days. 

I'm seeing a lot of changes that are just formatting, can you exclude these? It will make reviewing the actual changes much easier.

jewbetcha (changes requested):
Please exclude formatting changes

jewbetcha:
We've just now fixed this via identifying some drift with the OPS clock. Closing!

PR #85, opened 2026-05-12
Drop spin autocorr override and lower SNR floor

johnpacino:
## Summary

Two spin-detection fixes plus a debug script.

- **Drop the autocorr override branch in `detect_spin`.** The autocorrelation fallback was running for marginal-SNR captures (SNR in `[MIN, MEDIUM)`), and when the autocorr peak disagreed with the FFT peak but had correlation ≥ 0.4, it would *replace* the FFT result. In practice the autocorr search-region peak almost always lands at minimum lag (~12000 RPM, the upper rail) by spectral coincidence, so legitimate mid-range seam tones (e.g., 5000–7000 RPM 7-iron spin) were being flipped to the upper rail and then rejected as bandpass-shoulder noise. The fallback now only *confirms* the FFT pick when the two agree within 10%; disagreements are logged for diagnostics but never replace the FFT result. The post-autocorr rail re-checks become dead code and are removed.
- **Lower `SPIN_SNR_MIN` from 3.0 → 2.5.** Marginal but real seam tones are now reported at low confidence rather than dropped entirely.
- **Add `scripts/analysis/plot_spin_debug.py`.** Four-panel matplotlib figure for one `rolling_buffer_capture` entry: speed timeline, raw I/Q post-trigger, bandpass envelope, and the envelope FFT spectrum with leakage/rail zones and the picked peak annotated. Lets you see exactly what `detect_spin` saw and why it accepted or rejected a shot.

## Test plan

- [x] `uv run pytest tests/test_rolling_buffer.py` — 114 tests pass, including new `test_marginal_snr_preserves_fft_pick` regression test
- [x] `ruff check` clean on touched files
- [x] CHANGELOG entry added under `[Unreleased]`
- [ ] Manual: run plot script against a recent session and verify the 4 panels render with rejection reason annotated
- [ ] Manual: hit shots on Pi and confirm spin detection rate improves on captures that previously hit the autocorr override path

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,7 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Spin detection: drop the autocorrelation override branch. The autocorr
+  peak inside the envelope search region often lands at minimum lag
+  (~12000 RPM / upper rail) by spectral coincidence, which previously
+  flipped legitimate mid-range FFT seam picks to the upper rail and got
+  them rejected as bandpass-shoulder noise. The autocorr fallback still
+  *confirms* the FFT pick when the two agree within 10%; disagreements
+  are now logged for diagnostics but never replace the FFT result.
+- Spin detection: lower `SPIN_SNR_MIN` from 3.0 → 2.5 so marginal but
+  real seam tones are reported at low confidence instead of dropped.
+
 ### Added
+- `scripts/analysis/plot_spin_debug.py`: 4-panel diagnostic for a single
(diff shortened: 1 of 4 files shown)


johnpacino:
Like I mentioned, been playing around with spin settings and based on the changes you made earlier, i've narrowed it down to these two small changes. It took spin detection from single digits to 40%. 

Also the script generates a 4 table plot - that led me down a club speed idea that i'll submit another PR for - but visualizes the OPS data.

jewbetcha (approved):
Awesome, looking forward to trying this out

johnpacino:
I meant to add this previously but here's what the script outputs. 

shot12_spin

PR #86, opened 2026-05-12
Add club-speed median-picker replay analysis script

johnpacino:
## What this is

A diagnostic script — **not** a production code change. It runs a
proposed MEDIAN club-speed picker against any session log and reports
its result alongside the production (magnitude-pick) value for every
`rolling_buffer_capture` entry.

## Why

On a behind-the-ball setup, the production picker (highest candidate
magnitude) can land on the mid-swing beam-center reading rather than
on a reading closer to impact. In TrackMan comparison runs we saw a
discrepancy between reported club speeds and TM, and a few shots
produced smash factors above the physical iron/wedge maximum (~1.42).
Replaying captures through the proposed picker, the discrepancy
narrowed and smash factors fell back into realistic strike-quality
ranges.

## What the script does

1. Builds the same candidate set the production picker uses (outbound,
   67-85 % of ball speed, within 100 ms of ball, not the ball itself).
2. Applies a **magnitude floor**: keep candidates whose magnitude is
   ≥ 30 % of the peak candidate magnitude. Rejects sidelobe spikes.
3. Picks the **median speed** across surviving candidates. Per-shot
   inspection shows the candidate set is bimodal — a high-magnitude
   mid-swing cluster and a lower-magnitude approach cluster — and the
   median naturally lands near the boundary between them.

Run it on any session log:

```bash
.venv/bin/python scripts/analysis/replay_club_speed.py \
    session_logs/session_<timestamp>_range.jsonl
```

## What the production change would look like

If we move forward, the change in `find_club_speed` replaces the
magnitude pick at the end of the function with this:

```python
    if not candidates:
        return None, None

    # Reject sidelobe spikes by requiring magnitude >= 30% of the
    # peak candidate magnitude.
    peak_mag = max(r.magnitude for r in candidates)
    eligible = [r for r in candidates if r.magnitude >= peak_mag * 0.30]
    if not eligible:
        return None, None

    # Median candidate speed. The candidate set is bimodal (mid-swing
    # high-mag cluster + late-swing low-mag cluster bracket the
    # impact-speed reading); the median lands near the boundary.
    median = float(np.median([r.speed_mph for r in eligible]))
    pick = min(eligible, key=lambda r: abs(r.speed_mph - median))
    return median, pick.timestamp_ms
```

About 14 lines added, 4 lines removed in `find_club_speed`. Plus a
class constant for the 0.30 magnitude floor if we want it tunable.

## What this PR ships

- `scripts/analysis/replay_club_speed.py` — the diagnostic
- `docs/CHANGELOG.md` entry under \`[Unreleased]\`

No changes to \`processor.py\`, \`monitor.py\`, the UI, or anything in
production paths.

## Test plan

- [x] \`ruff check\` / \`ruff format\` clean on the new file
- [ ] Coleman to run the script against his own TrackMan-comparison
      sessions and decide whether MEDIAN's per-shot picks look more
      defensible than magnitude-pick

Related: #85

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -19,6 +19,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   real seam tones are reported at low confidence instead of dropped.
 
 ### Added
+- `scripts/analysis/replay_club_speed.py`: offline replay of a proposed
+  MEDIAN club-speed picker against any session log. Builds the same
+  candidate set the production picker uses, applies a 30 % magnitude
+  floor, and reports the median speed for each `rolling_buffer_capture`
+  alongside the originally logged (magnitude-pick) value, with smash
+  factors as a physical sanity check. The script is exploratory and
+  does not change production behaviour — it lets us inspect what a
+  median-based picker would have produced before committing to a code
+  change.
 - `scripts/analysis/plot_spin_debug.py`: 4-panel diagnostic for a single
   `rolling_buffer_capture` (speed timeline, raw I/Q, bandpass envelope,
   envelope FFT spectrum) to inspect what the spin algorithm saw and why
--- a/scripts/analysis/replay_club_speed.py
+++ b/scripts/analysis/replay_club_speed.py
@@ -1,0 +1,326 @@
+"""Replay a session log through a proposed MEDIAN club-speed picker and
+report before/after vs the logged (magnitude-pick) club speed.
+
+Background
+----------
+The production `find_club_speed` in `rolling_buffer/processor.py` picks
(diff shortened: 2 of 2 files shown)


johnpacino:
@jewbetcha - another one for you. This is just an offline script that you can use with session logs. With the plotting images created from the other script, i noticed that a lot of the shots that recorded lower than my trackman had the trackman club speed values in there but just not selected. The script uses a median methodology vs highest magnitude that can be thrown off depending on radar positioning. This improved my trackman comparison by a lot and also normalized smash factors to more realistic numbers.

If you're seeing closer club speeds with your trackman session, let me know and I can submit the switch in the code itself.

jewbetcha (approved):

PR #87, opened 2026-05-18
Add missing Flask server dependencies

jaceallison:
## What does this PR do?

Adds missing Flask server dependencies used by `src/openflight/server.py`.

The server imports these packages but they were not declared in `pyproject.toml`, which caused `make test` to fail during test collection on a fresh install:

- `flask`
- `flask-cors`
- `flask-socketio`

This PR adds those dependencies so a clean `make install` environment can import the server module correctly.

## How was this tested?

Tested on a fresh local setup with:

```bash
make install
make test


jewbetcha (approved):
Thank you!

PR #100, opened 2026-06-07
Fix OPS trigger timing with per-shot clock sync

johnpacino:
## Summary

This PR refreshes the OPS243 clock sync for each accepted sound-trigger capture instead of relying on the startup/previous sync indefinitely.

The stale OPS clock sync could drift during idle gaps, which shifted the K-LD7 impact timestamp by tens or hundreds of milliseconds. In testing, per-shot sync corrected stale-sync deltas as large as ~370ms while keeping the K-LD7 anchor within roughly one frame of impact.

## Changes

- Read a fresh OPS `C?` clock sync after a sound-trigger capture is accepted.
- Use the fresh sync when it passes quality checks.
- Fall back to a recent previous sync only when the fresh read is invalid.
- Fall back to first-byte timing only when no trustworthy clock sync is available.
- Add clock-sync quality gates for rollover uncertainty and timeout-like reads.
- Prevent diagnostic `read_clock_sync(store=False)` reads from replacing `last_clock_sync`.
- Add tests for fresh sync selection, previous-sync fallback, and non-mutating sync reads.

## Validation

- `uv run pytest tests/test_rolling_buffer.py tests/test_ops243.py -q`
- `uv run ruff check src/openflight/rolling_buffer/trigger.py src/openflight/ops243.py tests/test_rolling_buffer.py tests/test_ops243.py`

Manual validation:
- 9 spaced clap tests over a long-running session: all used fresh sync; anchor timing stayed about `-14ms` to `+13ms`.
- 25 golf shots: all used fresh sync; anchor timing stayed about `-16ms` to `+16ms`.

## Notes

This PR intentionally does not include K-LD7 debug endpoints, timing-analysis tooling, or the planned early K-LD7 snapshot architecture change.

### DIFF
--- a/src/openflight/ops243.py
+++ b/src/openflight/ops243.py
@@ -272,6 +272,7 @@ class OPS243Radar:
         per_read_timeout: float = 0.2,
         max_sync_duration_s: float = 1.25,
         sample_interval_s: float = 0.01,
+        store: bool = True,
     ) -> dict:
         """Map the OPS internal clock to host epoch via repeated ``C?`` reads.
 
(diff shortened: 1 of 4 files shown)


jewbetcha (approved):
Awesome

PR #102, opened 2026-06-08
fix: install kld7 driver by default so clean installs don't fail

aadnesd:
## Problem
On a clean install the K-LD7 angle radar fails with `kld7 package not installed`, forcing users to manually `pip install kld7`. `kld7` was an optional extra (`.[kld7]`), but no install path actually pulled it in:
- `scripts/setup/setup.sh` installs `.[ui,analysis]` — no `kld7`
- `make install` / `uv sync --group dev` don't install optional extras
- CI doesn't install it either

Only `start-kiosk.sh`'s error message mentioned `.[ui,kld7]`, so the extra was effectively never installed.

## Fix
Move `kld7>=0.2.1` into base `dependencies`. It's a 7KB pure-Python wheel whose only requirement (`pyserial`) is already a base dep, and the `--kld7` runtime flag still gates actual hardware use — so there's minimal downside to shipping it everywhere, and it can't be forgotten by any install path again. The existing defensive `find_spec("kld7")` import guards remain as a safety net.

- Removed the now-redundant `kld7` optional extra
- Updated the `start-kiosk.sh` hint to `.[ui]`
- Updated stale `pip install kld7` / `uv add kld7` messages in the kld7 driver and helper scripts to point at the project install (`uv pip install -e '.[ui]'`)
- Added a regression test asserting `kld7` is a base dependency

## Testing
- `uv run pytest tests/` → 581 passed, 6 skipped
- `uv run ruff check src/openflight/` → clean
- Confirmed `uv run` now auto-installs `kld7` and `import kld7` works

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ dependencies = [
     "flask>=3.1.3",
     "flask-cors>=6.0.2",
     "flask-socketio>=5.6.1",
+    "kld7>=0.2.1",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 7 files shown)


jewbetcha (commented):
Thank you!

jewbetcha (approved):

PR #104, opened 2026-06-09
fix(kiosk): launch server via uv run so a moved venv self-heals

pablitoarango:
## What does this PR do?

Makes `scripts/start-kiosk.sh` resilient to the project directory being moved.

I hit this firsthand: after moving my OpenFlight checkout to a different
directory, startup failed with a cryptic "command not found". I had Claude Code
diagnose it. The script was doing `source .venv/bin/activate`, but a virtualenv
bakes an **absolute** `VIRTUAL_ENV` path in at creation time — so once the repo
moved, `activate` pointed at a stale path and `openflight-server` resolved into a
`bin/` that no longer existed.

This PR resolves the environment dynamically instead of relying on a baked-in
venv path:

- Require `uv` up front (clear error + install link if it's missing).
- Run `uv sync --quiet` to reconcile `.venv` against `uv.lock`.
- Launch the server with `uv run $SERVER_CMD` instead of `$SERVER_CMD`.

**Why this is better than the current setup**

- **Self-healing**: a moved, stale, or missing `.venv` is rebuilt by `uv sync`
  rather than failing. The old flow required the venv to already exist *and* to
  sit at the exact path where it was created.
- **No baked-in absolute paths**: `uv run` resolves the entry point fresh each
  launch, so the repo is portable across directories.
- **Reproducible**: `uv sync` keeps the runtime env consistent with `uv.lock`,
  consistent with the project's "always use uv" rule.
- **Clearer failures**: a missing `uv` now gives an actionable message instead
  of a downstream "command not found".

## How was this tested?

Shell-only change. Verified `start-kiosk.sh` launches the server from a moved
checkout (the original failing scenario) and that a clean checkout with no
`.venv` now bootstraps via `uv sync` instead of erroring out.

## Checklist

- [ ] Python tests pass (`uv run pytest tests/ -v`) — N/A, no Python changed
- [ ] Pylint passes (`uv run pylint src/openflight/ --fail-under=9`) — N/A
- [ ] Ruff passes (`uv run ruff check src/openflight/`) — N/A
- [ ] UI builds (`cd ui && npm run build`) — N/A, no UI changed
- [ ] UI lint passes (`cd ui && npm run lint`) — N/A
- [x] Updated docs or CHANGELOG if needed — no docs reference the old activate flow
- [x] No unrelated changes mixed in — single-commit, single-file change


jewbetcha (approved):
Great, thanks for this 🚀

PR #106, opened 2026-06-10
Fix K-LD7 ring buffer race that silently dropped launch angles

dieterolson:
## What does this PR do?

Fixes a thread-safety race in `KLD7Tracker` that makes shots silently lose their launch angle.

The K-LD7 stream thread appends frames to the ring buffer (`_add_frame`, ~34 Hz) while the OPS243 shot path iterates the same `deque` — both `snapshot_buffer()` and `_radc_frames_for_extraction()` loop over `self._ring_buffer` with no synchronization. When an append lands mid-iteration, CPython raises `RuntimeError: deque mutated during iteration`. The server's K-LD7 exception handling then swallows it, so the symptom in the field is a shot that intermittently has no launch angle (or a missing buffer snapshot in the session log) with nothing in the logs to explain why. The race window exists on every shot while the stream is live.

The fix is minimal and changes no behavior otherwise:

- A `threading.Lock` is created alongside the ring buffer in `_init_ring_buffer()`.
- `_add_frame()` and `reset()` mutate the deque under the lock.
- `snapshot_buffer()` and `_radc_frames_for_extraction()` copy the deque under the lock and do all per-frame work (dict building, base64 encoding) on the copy, so lock hold time stays a single O(n) list copy of a ≤170-element buffer (~µs) and the stream thread is never blocked behind serialization work.

## How was this tested?

- **Failing test first** (per the repo's bug-fix rule): added `test_buffer_reads_are_safe_during_concurrent_appends`, which fills the buffer to `maxlen` (production steady state, every append evicts) and hammers both reader paths from the test thread while a writer thread appends. **Without the fix it fails in seconds with `RuntimeError: deque mutated during iteration` at the exact production call site (`snapshot_buffer`)**; with the fix it passes 5/5 consecutive runs.
- Full suite on Linux: all tests pass.
- Full suite on Windows: 611 passed; the only failures are 9 pre-existing environment-specific tests (bash-dependent `test_start_kiosk` tests and one CRLF-sensitive TrackMan CSV test) that fail identically on a clean checkout of `main`.
- `uv run pylint src/openflight/ --fail-under=9` → 9.69/10.
- `uv run ruff check` and `ruff format --check` clean on all touched files.
- `cd ui && npm run build && npm run lint` → both pass (no UI changes; verified anyway).

## Checklist


jewbetcha:
Sorry... accidental closure here. `git` history got in a weird state, can you re-open?

jewbetcha:
@dieterolson just want to make sure you see this, plz re-open when you get a chance!

dieterolson:
Working on it. It's saying most likely I can't reopen due to history changes - reopen button isn't showing up for me. Will try creating new PR based on the current state of the branch to retain the new commits.

dieterolson:
Pushed new PR 117

PR #107, opened 2026-06-13
fix(ops): don't truncate buffer dump when sound trigger fires near timeout

johnpacino:
## Problem

`wait_for_hardware_trigger()` bounds **total elapsed time**, not time-to-trigger. The 30s wait window and the dump readout share one deadline, so when an impact lands in the last ~5s of the window, the ~46KB rolling-buffer dump (~4-5s at serial speed) gets cut off mid-stream and the shot is silently lost.

At a 30s re-arm cadence this throws away roughly 1 in 6 randomly-timed shots.

## Observed

Live session, 2026-06-12 — trigger fired at 26.0s into the 30s window, read loop died at exactly 30.0s mid-I-array:

```
2026-06-12 09:22:19,438 - ops243 - DEBUG - [OPS] Hardware trigger: first byte after 26.0s
2026-06-12 09:22:23,463 - ops243 - INFO - [OPS] Hardware trigger: 40243 bytes in 30.0s
2026-06-12 09:22:23,463 - openflight.rolling_buffer.trigger - INFO - [TRIGGER] Sound trigger fired, 40243 bytes received
2026-06-12 09:22:24,217 - openflight.rolling_buffer.processor - WARNING - [PROCESSOR] Incomplete capture (missing: Q). Response (40243 bytes): '{"sample_time":946.077}...'
2026-06-12 09:22:24,218 - openflight.rolling_buffer.trigger - WARNING - [TRIGGER] Sound trigger parse failed (40243 bytes received)
```

A complete dump is ~46KB; 40,243 bytes in exactly 30.0s is the deadline killing the read, not a transfer problem.

## Fix

Once the first byte arrives, the trigger has fired and the dump is in flight — extend the deadline to `first_byte + dump_grace` (default 8s, ~2x the worst observed dump time). The existing completeness check (closed `Q` array) and 0.5s-inactivity exit are untouched, so a normal dump still returns the instant it completes; no latency is added to the common path.

- No data at all → still times out at the original `timeout`
- Trigger early in the window → behavior unchanged (deadline only ever extends, `max()`)

## Tests

Wrote the failing test first (`TestWaitForHardwareTrigger` in `tests/test_ops243.py`) using a `_ScheduledSerial` fake that releases bytes on a wall-clock schedule, scaled down (trigger at 0.3s of a 0.4s window, dump completes at 0.7s):

- `test_trigger_near_timeout_still_reads_full_dump` — reproduces the truncation against the old code, passes with the fix
- `test_no_trigger_returns_empty_after_timeout` — timeout path unchanged
- `test_complete_dump_returns_before_grace_expires` — grace adds no latency to a fast dump

Full suite: 636 passed. `ruff check` / `ruff format` clean.

### DIFF
--- a/src/openflight/ops243.py
+++ b/src/openflight/ops243.py
@@ -1051,7 +1051,7 @@ class OPS243Radar:
 
         return full_response
 
-    def wait_for_hardware_trigger(self, timeout: float = 30.0) -> str:
+    def wait_for_hardware_trigger(self, timeout: float = 30.0, dump_grace: float = 8.0) -> str:
         """
         Wait for hardware trigger to fire and read the buffer dump.
 
@@ -1060,7 +1060,11 @@ class OPS243Radar:
         (HOST_INT). Used with SoundTrigger (SparkFun SEN-14262).
 
         Args:
-            timeout: Maximum time to wait for trigger data
+            timeout: Maximum time to wait for the trigger to fire
+            dump_grace: Extra time allowed for the dump to finish once the
+                first byte has arrived. The ~46KB rolling-buffer dump takes
+                ~4-5s; a trigger firing near the end of the timeout window
+                must not have its dump cut off by the original deadline.
 
         Returns:
             Raw response string containing JSON lines, or empty string on timeout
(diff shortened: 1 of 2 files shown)


jewbetcha (approved):
Nice find

PR #113, opened 2026-06-15
Golf Simulator Connectors: stream shots to GSPro & OpenGolfSim

johnpacino:
## Summary

Adds an **optional, protocol-agnostic simulator-connector framework** so OpenFlight can stream shots into golf simulators in real time, with first-class support for **GSPro** (OpenConnect V1) and **OpenGolfSim**. The shot-detection pipeline is untouched — connectors are listeners that hang off the existing shot fan-out. Off by default; enabled with `--sim`.

## Architecture

A connector = a **codec** (one simulator's wire format) + a shared TCP **transport**. The pieces that are the same for every sim are written once:

| Layer | File | Shared? |
|---|---|---|
| TCP lifecycle, reconnect/backoff, JSON framing, optional heartbeat | `sim/transport.py` | shared |
| `Shot` → resolved fields + measured/estimated provenance (fallback table) | `sim/resolver.py` | shared |
| Connection/player state, inbound events, resolved shot | `sim/types.py` | shared |
| `SimConnector` + codec registry | `sim/codec.py` | shared |
| Config (`config/sim.json`) | `sim/config.py` | shared |
| **Wire format** (serialize shot, parse inbound, heartbeat) | `gspro/codec.py` (OpenConnect V1) | per-sim |

Adding a sim = one codec + a registry entry. The server holds a `list[SimConnector]` and fans each shot out to all connected sims (multiple at once supported).

## Connectors

Both ride the shared **OpenConnect V1** codec — they differ only in name + default port:
- **GSPro** — OpenConnect V1 on TCP 921.
- **OpenGolfSim** — its built-in **Developer API on TCP 3111** speaks OpenConnect V1; the connector reports as "OpenGolfSim".

## Enabling

Targets live in `config/sim.json` (`config/sim.example.json` provided). Connectors are **off unless `--sim`** is passed (e.g. `start-kiosk.sh --sim`); marked-`enabled` connectors then come up. (Default-on can come later once it's broadly stable.)

## UI

- One status pill per connector in the header (green/amber/red/gray).
- The club picker follows server-pushed club changes.
- A per-shot "Sent to <sim>" measured/estimated badge panel — shown only in debug mode.

## OpenGolfSim club sync (note)

OGS's Developer API returns an OpenConnect `201 Player` block but ships with a hardcoded `Club:"DR"` stub. `scripts/setup/opengolfsim/` carries a small local patch (+ `apply.sh`) that finishes the stub so club syncs; the durable fix is upstreaming that ~3-line change to OGS. Also: OGS rejects any `ShotNumber` it has already seen (a counter that persists across reconnects), so the shot counter is seeded from the clock to stay monotonic across restarts. Documented in `docs/simulator/opengolfsim.md`.

## Testing

- Python: full suite green (722 passed), `ruff` clean, `pylint` 9.4+ on the new packages.
- UI: 19 tests pass, production build clean.
- Connector behavior covered by unit + mock-TCP integration tests (framing, reconnect/backoff, heartbeat-optional, codec round-trips, resolver fallback table, server fan-out, concurrent connectors). Validated end-to-end against the live OpenGolfSim Developer API.

## Docs

`docs/simulator/` — architecture/pattern doc + per-sim setup guides (GSPro, OpenGolfSim).


jewbetcha:
Awesome. Can't wait to test this!

PR #115, opened 2026-06-16
Golf Simulator Connectors: stream shots to GSPro & OpenGolfSim

johnpacino:
> Re-opens the work from #113, which was accidentally closed. GitHub blocked reopening it (stale-state error on the closed PR), so this is a fresh PR on the same branch (`feat/sim-connectors`) with identical content.

## Summary

Adds an **optional, protocol-agnostic simulator-connector framework** so OpenFlight can stream shots into golf simulators in real time, with first-class support for **GSPro** (OpenConnect V1) and **OpenGolfSim**. The shot-detection pipeline is untouched — connectors are listeners that hang off the existing shot fan-out. Off by default; enabled with `--sim`.

## Architecture

A connector = a **codec** (one simulator's wire format) + a shared TCP **transport**. The pieces that are the same for every sim are written once:

| Layer | File | Shared? |
|---|---|---|
| TCP lifecycle, reconnect/backoff, JSON framing, optional heartbeat | `sim/transport.py` | shared |
| `Shot` → resolved fields + measured/estimated provenance (fallback table) | `sim/resolver.py` | shared |
| Connection/player state, inbound events, resolved shot | `sim/types.py` | shared |
| `SimConnector` + codec registry | `sim/codec.py` | shared |
| Config (`config/sim.json`) | `sim/config.py` | shared |
| **Wire format** (serialize shot, parse inbound, heartbeat) | `gspro/codec.py` (OpenConnect V1) | per-sim |

Adding a sim = one codec + a registry entry. The server holds a `list[SimConnector]` and fans each shot out to all connected sims (multiple at once supported).

## Connectors

Both ride the shared **OpenConnect V1** codec — they differ only in name + default port:
- **GSPro** — OpenConnect V1 on TCP 921.
- **OpenGolfSim** — its built-in **Developer API on TCP 3111** speaks OpenConnect V1; the connector reports as "OpenGolfSim".

## Enabling

Targets live in `config/sim.json` (`config/sim.example.json` provided). Connectors are **off unless `--sim`** is passed (e.g. `start-kiosk.sh --sim`); marked-`enabled` connectors then come up. (Default-on can come later once it's broadly stable.)

## UI

- One status pill per connector in the header (green/amber/red/gray).
- The club picker follows server-pushed club changes.
- A per-shot "Sent to <sim>" measured/estimated badge panel — shown only in debug mode.

## OpenGolfSim club sync (note)

OGS's Developer API returns an OpenConnect `201 Player` block but ships with a hardcoded `Club:"DR"` stub. `scripts/setup/opengolfsim/` carries a small local patch (+ `apply.sh`) that finishes the stub so club syncs; the durable fix is upstreaming that ~3-line change to OGS. Also: OGS rejects any `ShotNumber` it has already seen (a counter that persists across reconnects), so the shot counter is seeded from the clock to stay monotonic across restarts. Documented in `docs/simulator/opengolfsim.md`.

## Testing

- Python: full suite green (722 passed), `ruff` clean, `pylint` 9.4+ on the new packages.
- UI: 19 tests pass, production build clean.
- Connector behavior covered by unit + mock-TCP integration tests (framing, reconnect/backoff, heartbeat-optional, codec round-trips, resolver fallback table, server fan-out, concurrent connectors). Validated end-to-end against the live OpenGolfSim Developer API.

## Docs

`docs/simulator/` — architecture/pattern doc + per-sim setup guides (GSPro, OpenGolfSim).


src/openflight/server.py
```diff
@@ -1485,6 +1510,134 @@ def handle_shutdown():
...
+        socketio.emit("sim_shot_dropped", {"reason": str(e)})
+        return
+
+    values = resolved.as_values()
+    for connector in sim_connectors:
+        if not connector.is_connected():
+            continue
+        try:
+            connector.send_shot(resolved)
+        except OSError as e:
```
jewbetcha, line 1552: **(High — likely merge blocker) This `except OSError` does not cover the failure it's guarding against, so a sim disconnect mid-shot can raise into the shot pipeline.**

`_forward_shot_to_simulators` documents *"never raises into it"* and guards each send with `is_connected()`. But `send_shot` → `TcpSimClient.send_raw` (transport.py:166) raises **`RuntimeError`**, not `OSError`, when the socket has gone away:

```python
def send_raw(self, data):
    with self._sock_lock:
        if self._sock is None:
            raise RuntimeError("send_raw called while not connected")
```

There is a real TOCTOU window: in `_connection_loop`, when a connection drops, `_close_socket()` sets `_sock = None` **before** `_set_state(RECONNECT_BACKOFF)`. So between `is_connected()` returning `True` here and `send_shot` acquiring `_sock_lock`, `_sock` can become `None` → `RuntimeError` propagates up through `on_shot_detected`. That's exactly the disruption the docstring promises can't happen, and it would fire precisely when a sim drops while a shot is being processed.

**Suggested change** (minimal, keeps the guarantee honest):
```python
        except (OSError, RuntimeError) as e:
            logger.warning("[sim] %s send failed: %s", connector.name, e)
            socketio.emit("sim_send_failed", {"target": connector.name, "reason": str(e)})
            continue
```

Alternative (cleaner contract, larger blast radius): change `send_raw` to raise `ConnectionError` (an `OSError` subclass) instead of `RuntimeError`, and keep this catch as `OSError`. Either works; the catch-both is the smaller diff. Per the repo's bug-first rule, a test that closes the socket between `is_connected()` and `send_shot` and asserts the pipeline doesn't raise would lock this down.

src/openflight/server.py
```diff
@@ -1485,6 +1510,134 @@ def handle_shutdown():
...
+        logger.warning("[sim] shot not sendable: %s", e)
+        socketio.emit("sim_shot_dropped", {"reason": str(e)})
+        return
+
+    values = resolved.as_values()
+    for connector in sim_connectors:
+        if not connector.is_connected():
+            continue
+        try:
+            connector.send_shot(resolved)
```
jewbetcha, line 1551: **(Low — design note, not a required change) Network send runs synchronously on the shot-detection thread.**

`send_shot` is a blocking `sendall`, and the socket's timeout is whatever `_recv_loop` last set (0.2s). With a stalled sim TCP send buffer, this adds up to ~0.2s per connected connector of latency to `on_shot_detected`. Negligible for localhost sims, so no change needed now — but worth a one-line comment acknowledging the coupling, and a candidate for a per-connector send queue if remote/slow sims ever land.

src/openflight/sim/transport.py
```diff
@@ -0,0 +1,319 @@
...
+            sock.settimeout(0.2)
+            try:
+                data = sock.recv(4096)
+            except socket.timeout:
+                continue
+            except OSError:
+                return
+            if not data:
+                return
+            buffer.extend(data)
```
jewbetcha, line 311: **(Medium) Unbounded receive buffer — a frame that never closes its braces grows memory without limit.**

`find_json_end` returns `None` until a top-level `{...}` is balanced. If a sim ever sends data whose braces never close (truncation, a protocol bug, or a hostile peer), `buffer.extend(data)` here grows forever — there is no max-frame backstop, and `find_json_end` re-scans the whole buffer each `recv`. Localhost + trusted sim makes this low-likelihood, but it's an unbounded-growth path with no guard.

**Suggested change:** cap the buffer and reset on overflow (frames are tiny — a few hundred bytes):
```python
_MAX_FRAME_BYTES = 64 * 1024
...
            buffer.extend(data)
            if len(buffer) > _MAX_FRAME_BYTES:
                logger.warning("[%s] inbound buffer exceeded %d bytes with no complete frame; resetting",
                               self._name, _MAX_FRAME_BYTES)
                buffer.clear()
                continue
```

src/openflight/sim/config.py
```diff
@@ -0,0 +1,73 @@
...
+
+def load_sim_config(config_path: Path = DEFAULT_CONFIG_PATH) -> List[ConnectorConfig]:
+    """Resolve the enabled connector configs from the file (only enabled ones).
+
+    Gating the whole feature on/off is the caller's job (the ``--sim`` flag);
+    this just reads which connectors the file enables.
+    """
+    if not config_path.exists():
+        return []
+    data = json.loads(config_path.read_text(encoding="utf-8"))
```
jewbetcha: **(Medium) A malformed `config/sim.json` crashes server startup for what is an opt-in, secondary feature.**

`json.loads(config_path.read_text(...))` has no error handling, so a syntactically broken sim config raises `JSONDecodeError` straight out of `main()` and takes down the entire launch monitor — even though sim is enabled only via `--sim` and the core shot pipeline doesn't depend on it. (`_with_defaults` has the same exposure on a non-int `port`.)

**Suggested change:** degrade to "no connectors" with a warning rather than aborting startup:
```python
    try:
        data = json.loads(config_path.read_text(encoding="utf-8"))
    except (json.JSONDecodeError, OSError) as e:
        logger.warning("[sim] ignoring unreadable %s: %s", config_path, e)
        return []
```
(Keep the existing `ValueError` raise for an *unknown connector type* — that's a real misconfiguration worth surfacing loudly. This is specifically about not letting a typo in an optional file brick the monitor.)

src/openflight/sim/types.py
```diff
@@ -0,0 +1,134 @@
...
+
+    handed: str = "RH"
+    club: ClubType = ClubType.DRIVER
+    shot_counter: int = 0
+
+    def next_shot_number(self) -> int:
+        self.shot_counter += 1
+        return self.shot_counter
+
+    def apply(self, update: PlayerUpdate) -> None:
```
jewbetcha, line 141: **(Low–Medium) `PlayerState` is shared mutable state across threads with no synchronization, and concurrent connectors are a supported feature.**

`apply()` runs on each connector's recv thread (`_sim_on_inbound`), while `next_shot_number()` runs on the shot thread. With multiple connectors enabled, two recv threads can call `apply()` concurrently, and `club`/`handed`/`shot_counter` are read/written cross-thread. CPython's GIL makes the `+= 1` and attribute writes incidentally safe today, but this is unsynchronized-by-design state for a path the PR explicitly supports (multiple sims at once).

**Suggestion:** add a `threading.Lock` (or `threading.RLock`) and take it in `apply()` and `next_shot_number()`. Low effort, makes the concurrency correct rather than relying on GIL semantics. Not a blocker.

src/openflight/sim/resolver.py
```diff
@@ -0,0 +1,138 @@
...
+    back_spin = total_spin * math.cos(axis_rad)
+    side_spin = total_spin * math.sin(axis_rad)
+    derived_prov = (
+        "measured" if (spin_prov == "measured" and axis_prov == "measured") else "estimated"
+    )
+    provenance["back_spin"] = derived_prov
+    provenance["side_spin"] = derived_prov
+
+    carry = float(shot.estimated_carry_yards)
+    provenance["carry"] = "measured" if shot.has_launch_angle else "estimated"
```
jewbetcha, line 112: **(Minor) `carry` provenance is labeled `"measured"`, but carry is never directly measured.**

The value is always `shot.estimated_carry_yards` (line 107) — an estimate, refined by measured launch angle when available. Tagging the badge `"measured"` when `has_launch_angle` slightly overstates it to the user. Not wrong as a UI heuristic, but consider a brief comment clarifying intent ("carry is modeled; 'measured' here means launch-angle-informed vs. fully defaulted"), or relabel to something like `"derived"` if the UI can express a third state.

src/openflight/server.py
```diff
@@ -25,6 +25,18 @@
 from .ops243 import Direction, SpeedReading, set_show_raw_readings
 from .rolling_buffer.monitor import estimate_carry_with_spin, get_optimal_spin_for_ball_speed
 from .session_logger import get_session_logger, init_session_logger, log_session_error
+from .sim import (
```
jewbetcha, line 28: **(Trivial) Two consecutive `from .sim import (...)` blocks** — the second only aliases `PlayerState as SimPlayerState`. These can be folded into a single import statement. Cosmetic; flagging only because it may trip isort/ruff import grouping.

jewbetcha (commented):
Reviewed the connector framework — the codec/transport/resolver split is clean, the shot pipeline is genuinely untouched, and the architecture makes a third sim a one-codec change as advertised. A few items worth addressing before merge, inline. Severity in each comment; #1 is the only one I'd treat as a blocker.

Nothing here is a nitpick about the design — the structure is good. These are correctness/robustness edges, weighted toward the failure paths.

src/openflight/server.py
johnpacino: Good catch — this was a real hole. Fixed in 45135ee: `send_raw` now raises `ConnectionError` (an `OSError` subclass) instead of `RuntimeError` when `_sock` is `None`, so the `except OSError` guard in `_forward_shot_to_simulators` actually catches a send that races a disconnect. Added `test_send_raw_while_disconnected_raises_oserror_subclass` for the raise itself, plus `test_forward_swallows_send_failure` proving the shot pipeline doesn't raise when one connector drops mid-fan-out (and the other still gets the shot).

src/openflight/server.py
johnpacino: Acknowledged — no functional change. Added a comment at the send site (4c2f132) recording that sends are synchronous on the shot thread in connector order, and that the fix, if a remote or laggy sim is ever added, is a per-connector send queue so one slow sim can't stall delivery to the others. Happy to build the queue now if you'd rather, but it felt like premature infrastructure while both targets are local and ack in microseconds.

src/openflight/sim/transport.py
johnpacino: Fixed in 5f97a91. Added a 64KB backstop in `_recv_loop`: after draining all complete frames, if the leftover incomplete buffer exceeds the cap we log a warning and clear it instead of growing without bound. `test_recv_buffer_resets_on_oversized_unclosed_frame` feeds a never-closing frame followed by a valid one and asserts the valid frame still parses — i.e. we resync rather than wedge behind the bad frame.

src/openflight/sim/config.py
johnpacino: Fixed in 83c7f5a. `load_sim_config` now degrades to "no connectors" with a warning on an unreadable/broken file (sim is opt-in; the core shot pipeline doesn't depend on it), and skips a single malformed connector entry so it can't take the others down with it. An *unknown connector type* still raises — that's a real misconfig worth surfacing loudly. Added tests for broken JSON, wrong top-level shape, an unreadable path, and one-bad-entry-others-kept.

src/openflight/sim/types.py
johnpacino: Fixed in c6ad283 — added a `threading.Lock` guarding `next_shot_number()` and `apply()`. You're right that the GIL makes it incidentally safe today: I tried fairly hard to reproduce a lost increment (8 threads, up to ~1.6M calls, switch interval down to 1e-9) and couldn't on CPython. So the two added tests are honest *guarantee* tests — they pin the invariants (a unique number per call; an all-or-nothing field update) and would catch a regression on a free-threaded build, rather than reproducing a failure today. Called that out in the commit message so it's not mistaken for a red→green.

src/openflight/sim/resolver.py
johnpacino: You're right that carry is never directly measured. Rather than relabel (which would change the UI badge semantics), I added a comment (4c2f132) clarifying that "measured" here means *launch-angle-informed* — the carry model was driven by a measured launch angle rather than club-type defaults — not a claim that carry itself was observed. If you'd prefer a distinct provenance value (e.g. `"derived"`) so the UI can show a third state, I can do that as a follow-up.

src/openflight/server.py
johnpacino: Folded into one block in 4c2f132. The split was actually ruff-enforced — its default `combine-as-imports = false` puts the aliased member in its own statement — so I enabled `combine-as-imports = true` under `[tool.ruff.lint.isort]`. Verified it only affects this one block: every other aliased import in `src` is a plain module alias (`numpy as np`), which the setting doesn't touch.

jewbetcha (approved):

PR #118, opened 2026-06-16
fix: make SessionLogger writes thread-safe

dieterolson:
## What does this PR do?

Makes `SessionLogger` writes thread-safe.

The `log_*` methods all funnel into `_write_entry`, which writes one line to a single session JSONL file. In production that file is written from several threads at once — the OPS243 capture thread (`log_shot`, `log_rolling_buffer_capture`), the K-LD7 stream thread (`log_session_error`), and Flask-SocketIO handlers (`log_config_change`). `_write_entry` held no lock, which allowed two failure modes under load:

1. **Interleaved/corrupt JSONL lines.** Large entries exceed the underlying text stream's atomic-write size — `rolling_buffer_capture` carries 2×4096 raw samples and `kld7_buffer` carries base64 RADC payloads. Concurrent writes can interleave mid-line. Those logs are the project's offline replay corpus, so a single corrupt line silently breaks post-session analysis.
2. **Write racing close.** `end_session()` closes the file and sets it to `None` while another thread is mid-write, raising `AttributeError` / `ValueError: I/O operation on closed file`.

The fix adds a `threading.Lock` that serializes `_write_entry`'s `write`+`flush` and `end_session`'s close. The JSON is serialized *outside* the lock so the critical section is just the write+flush and the stream thread is never blocked behind another thread's encoding work. This mirrors the locking approach used elsewhere for shared buffers.

**Scope note (intentional):** the per-shot stat counters (`self._stats[...] += 1`) are left unsynchronized — shots are produced serially by the capture thread, so they are not a corruption vector, and widening the lock there is out of scope for this focused fix.

## Why was this required?

The session JSONL files are the project's offline replay corpus — every post-session analysis, spin-detection review, and TrackMan comparison reads them back. Because `_write_entry` was called concurrently from three thread contexts with no lock, two latent bugs could silently damage that corpus in the field:

- A large `rolling_buffer_capture` or `kld7_buffer` line interleaving with another thread's write produces a corrupt JSONL line. The corruption is silent at capture time and only surfaces later when offline tooling fails to parse the session — after the range session is over and unrepeatable.
- A write racing `end_session()` closing the file raises `AttributeError` / `ValueError: I/O operation on closed file`, which can lose the tail of a session (including the `session_end` summary).

Both are data-loss bugs on irreplaceable field data, which is why serializing the writer is required rather than cosmetic.

## Automated tests

Following the repo's bug-fix rule, two deterministic regression tests were added to `tests/test_session_logger.py`, both of which **fail on the current code and pass with the fix**:

- `test_concurrent_writes_do_not_overlap_or_drop_entries` drives 8 threads through `_write_entry` against an instrumented stream whose `write()` flags any overlapping entry; without the lock the overlap is detected on every run, and it also asserts all entries are written exactly once as intact JSON lines.
- `test_end_session_does_not_close_during_an_active_write` blocks a write in flight and asserts `end_session()` does not close the file until that write completes. On the unfixed code this reproduces the use-after-close (`'NoneType' object has no attribute 'flush'`).
- Both were run 5× consecutively to confirm they are not flaky.

Full suite:
- Linux: **621 passed**, 6 skipped.
- `uv run pylint src/openflight/ --fail-under=9` → 9.69/10.
- `uv run ruff check` and `ruff format --check` clean on the touched files.
- `cd ui && npm run build && npm run lint` → both pass (no UI changes; verified for the checklist).

## Manual (human) testing

- Ran `uv run pytest tests/test_session_logger.py -v` by hand and watched the two new concurrency tests: confirmed they **fail on `main`** (overlap detected every run; `end_session` closes mid-write) and **pass with the fix applied**, then re-ran them 5× back-to-back to confirm they are not flaky.
- After resolving the merge with `main` (which independently added the `session_uuid`/cloud-sync fields to the same file), re-ran the full `test_session_logger.py` suite (23 tests) locally and confirmed all pass, including main's `TestSessionIdentity` tests alongside the new `TestWriteEntryThreadSafety` tests.
- Verified by hand that the JSON is serialized outside the lock (the critical section is only `write`+`flush`) by reading through `_write_entry`, so the K-LD7 stream thread is never blocked behind another thread's `json.dumps`.

## Checklist

- [x] **Single feature/fix** — scoped to making `SessionLogger` writes thread-safe
- [x] **Automated tests included** — two new regression tests in `tests/test_session_logger.py`
- [x] **Manual testing described** — documented above


jewbetcha (approved):
Nice 👌

jewbetcha:
@dieterolson still planned?

dieterolson:
Sorry - in process of moving and my computers are down for a bit. Haven't had much chance to code. Will follow up in a bit.

dieterolson:
Rebased onto current main (9b61f0a) — same single commit, +161/−8, no functional delta from the approved version. Conflicts were adjacency only: the import block (`uuid` on main vs `threading` here), both branches appending test classes to `test_session_logger.py`, and the CHANGELOG entry (now folded into the existing `### Fixed` section).

Re-verified after rebase: `pytest tests/test_session_logger.py` → 26 passed (incl. the two concurrency tests); full suite → 1138 passed with only the pre-existing Windows-environment failures, none new; `ruff check` clean; `pylint src/openflight/ --fail-under=9` → 9.70. Happy to re-verify anything else before merge.

PR #120, opened 2026-06-17
Redesign the kiosk UI + TV/web remote (glass, i18n, view modes, history, remote)

johnragonhall:
## What does this PR do?

Full redesign of the React kiosk UI for the Pi touchscreen and TV mode, plus the TV-remote / web-remote control story:

- glass design-token system and theme (color-blind remap + responsive root scale ramp)
- in-app localization across 16 languages
- interactive / scoreboard / phone-remote view modes with D-pad spatial navigation and the focus-ring system
- redesigned shot display (gauges, metrics, spin provenance), plus stats analysis views and shot charts
- session history view backed by `/api/history` endpoints
- unit preferences, refreshed core components, and the app shell + socket wiring
- web remote: `/scoreboard` and `/remote` server routes with a D-pad key relay, view-mode autodetect at startup, and an HDMI-CEC / FLIRC IR bridge mapping a TV remote to kiosk navigation
- cloud sync: a header badge mirroring the connection status (color-blind / grayscale safe) plus a kiosk settings toggle to pause/resume uploads, backed by `get_cloud_status` / `set_cloud_sync` Socket.IO events the server gates to localhost
- club-select fix: pin the close X so the focus ring can't shift it, gate the TV focus ring behind real remote/keyboard nav, and drop the gold pre-highlight

Split into granular commits by area (BACKEND, DESIGN, UI, DOCS, BUILD).

## How was this tested?

- `cd ui && npm run build` clean (tsc + vite)
- `cd ui && npm run lint` clean
- `cd ui && npx vitest run` passes
- `uv run pytest tests/test_server.py` (93 passed, 1 skipped)
- `uv run ruff check` clean
- `tests/test_start_kiosk.py` shells out to `bash start-kiosk.sh --dry-run`, so it runs on the Linux CI rather than the Windows dev box

## Checklist

- [x] Python tests pass (`test_server`; `start-kiosk` tests are bash/Linux-only)
- [x] Ruff passes
- [x] Updated docs (user guide, TV-remote control)


jewbetcha:
Could you please add some before/after screenshots to the PR description?

jewbetcha:
I'm also seeing a lot of new html files, images, and AI files. Please exclude any ai files and images if they aren't used in the code. 

What's the story behind the html?

What about the QR code components? 

This is a huge PR with many different changes. I'm unlikely to merge this as is unless there's a cohesive story.

jewbetcha:
@johnragonhall I need you to explain these changes more. this is a 22k line change, far too big. 

> in-app localization across 16 languages

this alone could be its own PR. I need tests for all of this and you need to manually test yourself. 

While I am a supporter of AI tools, this is too heavy-handed for an OSS project and has far too big of a blast radius.

johnragonhall:
Thanks for the thorough review, Coleman. You're right on all counts. Here's what I've done and what's left to address.

---

**Split into smaller PRs**

The pieces that are genuinely self-contained have been pulled out:

- **#125** — i18n only: the translation catalog and language provider. Zero dependencies. 4 files.
- **#126** — Design tokens only: `tokens.css`, `App.css`/`index.css` updates, `useThemeColors`. Zero dependencies. 4 files.
- **#127** — Backend only: session history API (`/api/history`), scoreboard/remote server routes, TV remote bridge scripts (`cec_remote.py`, `flirc_setup.sh`), `start-kiosk.sh` updates. Zero UI coupling.

What remains in this PR is the core UI rewrite: shot display, view modes, spatial navigation, stats, and the App shell that wires it together. Those components share types and utilities too tightly to pull apart without breaking the TypeScript build (the shot display commit removes `useShotContext.ts` which the old App.tsx imports — they have to move together).

---

**Generated docs removed**

`docs/user-guide.html`, `docs/user-guide.md`, `docs/shot-visualization.md`, and `docs/tv-remote-control.md` are gone from this branch. The HTML file was a generated duplicate of the markdown. The markdown files will come back in a separate docs-only PR once the code is merged and the docs can be validated against the real implementation.

---

**The HTML files**

`ui/index.html` is the Vite app entry point. Every Vite project has it; the build fails without it. It's not extra content.

`docs/user-guide.html` is removed (see above).

---

**The QR code component**

`PairingQR.tsx` renders the pairing QR code for the BLE mobile companion (PR #122). Its backend endpoint is `/api/pair-qr`, which lands in #122. The component renders nothing until that endpoint exists. I've left it here because it lives in the UI layer and `SettingsPanel.tsx` imports it — moving it to #122 would split the settings panel across two PRs. Happy to stub it out behind a feature check if you'd prefer.

---

**Tests and manual verification**

You asked for both. I'll add:

1. A translations completeness test that asserts every key in `translations.ts` has a non-empty value for every language.
2. Before/after screenshots of the shot display, history view, scoreboard mode, and remote controller — running against the mock server so no hardware needed.

I'll push those to this branch before asking for another look.

---

**What to review now**

If you want to start anywhere, #125 (i18n) and #126 (design tokens) are the smallest and have no dependencies. #127 (backend) is Python-only and fully tested. This PR (#120) is the last to merge — it depends on the others being in first.

johnragonhall:
Ok let me reorganize this PR into smaller pieces the html is just a visual user guide to to all the new features. Some of the files can be split up and seperate PRs because they are new files. But essentially I tried to explore what a companion app would look like and then realized the previous UI had no design-token layer. 

This PR introduces a CSS property system where there single source of truth for every color, radius, shadow, and spacing unit. The token system includes a color-blind remap and a responsive root scale ramp so the UI reflows correctly from a 7-inch Pi touchscreen to a TV without manual breakpoints. I also made every user-visible string route through a translation layer. The initial release ships 16 languages. This make it so adding any language is just  a simple update. There are new view modes one is an interactive mode designed to fit the 7"  screen up to a 98" 8k TV. It has a live stats view of metrics from the radar. I designed the live tab to have easy to read numbers and feedback on shots. The UI also detects if it is a tv a monitor or the Pi for a low power mode. It also adds a second view mode (scorebaord) where it just shows live ball arcs and distances on a graphs and numbers designed for like a garage setup and some one who wants and extra screen to visualize stats. I also added ways to conrtol through tv remots voiceover labels for tvs and a phone remote where on any browser on the local network some one can use their phone as a D-pad controller. A user can change views or settings during a session without walking to the screen between shots.

The kiosk detects at startup whether the attached display is a TV (via EDID physical size and CEC presence) and seeds the appropriate mode. The user's saved choice always overrides the detection. Every interactive element in the UI supports keyboard and remote navigation. useSpatialNavigation tracks whether the current input method is a pointer or a D-pad, sets html.using-dpad during remote use, and clears it on the next pointer event. There is a gold focus ring appears only during remote navigation. Mouse and touch users never see the ring. 

The shot gauges, metric cards, and spin indicators were rebuilt. Spin sources (estimated vs radar-measured) are labeled per-shot so users know what the system is reporting and how confident it is. The angle source on club AoA and club path cards (geometry model vs direct radar) is surfaced the same way.

A session list view backed by the /api/history REST endpoints lets users scroll back through previous rounds on the kiosk. Individual shots within a session are expandable. No additional storage is required; the JSONL session files already on disk power the API.

/scoreboard and /remote are new server routes. /remote accepts D-pad key events from any browser on the LAN and relays them to the kiosk. The server also maps HDMI-CEC and FLIRC IR signals to the same key events, so a TV remote or a FLIRC-programmed universal remote drives the kiosk without any extra hardware or configuration.

A header badge mirrors the cloud sync connection state in real time. The badge uses color-blind-safe and grayscale-safe encoding. A settings toggle in the kiosk lets the operator pause or resume cloud uploads without restarting the server. The toggle calls set_cloud_sync over Socket.IO; the server gates that event to localhost so only the kiosk itself can change it.

I tested it in mock mode un a 7 inch screen and on a laptop and a tv. All the UI works and all the new settings and fetaures work as well. I have tested keybaord, remote and touch naviagtion and used Apple HIG as a reference for standards. The only feature I havent tested is all the voirceover labels but everyhting else leikt he translations works.

PR #124, opened 2026-06-17
refactor(ui): migrate UI state to Zustand and extract socket service

clschnei:
## What does this PR do?

Refactors the React UI state layer from React context/providers to Zustand stores, and extracts socket lifecycle/event handling into a dedicated service.

Main changes:
- add Zustand as a UI dependency
- replace shot, unit preference, LaunchDaddy, camera, debug, and system context state with focused stores
- move socket event listeners and emitters out of `useSocket` into `socketService`
- preserve newer simulator connector UI/state (`simStatuses`, `latestSimShots`, `serverClub`) while moving that state into the Zustand-backed app model
- simplify `App.tsx` so it reads from stores and calls the socket service directly
- move socket- and simulator-specific TypeScript types into `ui/src/types/socket.ts`

## Why was this required?

The UI had accumulated multiple context/provider layers plus a large transport-aware `useSocket` hook that owned most app state. That made the app harder to reason about, forced provider nesting in `App.tsx`, and mixed socket transport concerns with UI state updates.

This refactor separates responsibilities more cleanly:
- Zustand owns shared UI/application state
- `socketService` owns socket lifecycle and event wiring
- components read focused state from stores instead of depending on a monolithic hook

This also made it possible to retain recent upstream simulator connector work while putting its state (`simStatuses`, `latestSimShots`, `serverClub`) into the same canonical state model.

## Automated tests

Automated checks run for this branch:
- `UV_CACHE_DIR=/tmp/uv-cache TMPDIR=/tmp uv run --no-sync pytest tests/ -v` -> `851 passed, 6 skipped`
- `UV_CACHE_DIR=/tmp/uv-cache TMPDIR=/tmp uv run --no-sync pylint src/openflight/ --fail-under=9` -> `9.49/10`
- `UV_CACHE_DIR=/tmp/uv-cache TMPDIR=/tmp uv run --no-sync ruff check src/openflight/`
- `cd ui && npm run build`
- `cd ui && npm run lint`
- `cd ui && npm test` -> `6 files, 19 tests passed`
- `cd ui && PATH=/home/lev/Code/openflight/.venv/bin:$PATH CI=1 npm run test:e2e` -> `6 passed`

Notes:
- some commands needed writable temp/cache directories in this environment
- Python socket-using tests and Playwright e2e needed an unrestricted run because they open local sockets/ports

## Manual (human) testing

Manual hardware testing was not performed in this environment.

What I did verify directly while working:
- inspected the merged/rebased UI state flow after rebasing onto `upstream/main`, specifically preserving simulator connector UI/state (`SimStatus`, `SimShotBadges`, and server-driven club changes) alongside the Zustand refactor
- verified the app still builds cleanly and the Playwright e2e suite passes against the mock backend/UI stack
- reviewed the live/debug/camera code paths after the rebase conflict resolution to ensure socket-service calls and selected store reads still line up with the rendered views

Not manually verified here:
- real hardware flows
- live browser interaction outside Playwright
- camera streaming on actual hardware

## Checklist

- [ ] Updated docs or CHANGELOG if needed


ui/src/App.tsx
```diff
@@ -61,27 +62,13 @@ const Icons = {
...
-    updateRadarConfig,
-    toggleCamera,
-    toggleCameraStream,
-    shutdown,
-  } = useSocket();
+  const { shutdown } = useSocket();
+  const { connected, mockMode, debugMode } = useSystemStore();
+  const { latestShot, shots, isNewShot, shotVersion } = useShotStore();
+  const { cameraStatus } = useCameraStore();
+  const { debugReadings, debugShotLogs, radarConfig, triggerDiagnostics, triggerStatus } = useDebugStore();
```
jewbetcha: **Use selectors instead of whole-store destructuring.** Calling a Zustand hook with no selector subscribes to the *entire* store, so this whole component re-renders on any change in each store. `useDebugStore()` is the worst case: `debug_reading` / `trigger_diagnostic` stream at high frequency, so every one of them now re-renders the full `App` tree even when the debug view isn't mounted.

This isn't a regression (the old `useSocket` held all of it in one component's state and re-rendered just as often), but the refactor is the natural moment to scope it. Idiomatic form:

```ts
const connected = useSystemStore((s) => s.connected);
const mockMode = useSystemStore((s) => s.mockMode);
// or group with useShallow:
const { debugReadings, radarConfig } = useDebugStore(useShallow((s) => ({
  debugReadings: s.debugReadings,
  radarConfig: s.radarConfig,
})));
```

Keeps each view subscribed only to what it actually reads.

jewbetcha (commented):
Reviewed the Zustand/socketService refactor — the mechanical translation looks faithful (trimming caps, `isNewShot` timer, LaunchDaddy tap state machine all preserved) and nothing reads like a correctness regression. A few things worth addressing, plus two non-line-specific notes:

**Tests:** Real logic moved into testable units (shot-store debounce timer, debug-store ring-buffer trimming, `updateTriggerStatusStats` accumulation, LaunchDaddy tap window) but the only test change is removing a provider wrapper. These stores are now trivial to unit-test without a React render — given the project's testing bar and the long list of manual checks marked not-run, this is the main gap I'd want closed before merge.

**Type-home consistency:** `DebugReading`/`RadarConfig`/`DebugShotLog` moved to `types/socket.ts`, but `CameraStatus` is still exported from `stores/useCameraStore.ts` and imported from there by components. Pick one convention.

Inline comments below.

ui/src/hooks/useSocket.ts
```diff
@@ -1,274 +1,23 @@
...
-
-    socketRef.current = newSocket;
+    socketService.connect();
 
     return () => {
-      newSocket.close();
-      socketRef.current = null;
+      // Depending on if you want the socket to stay alive across remounts
+      // you might not want to disconnect here, but for now we'll match original behavior.
+      socketService.disconnect();
```
jewbetcha: `connect()` guards against double-connect, but `disconnect()` unconditionally closes the shared socket. Since `socketService` is a module singleton, if `useSocket` is ever called from a second component the first one to unmount tears down the socket for everyone. Safe today (only `AppContent` uses it), but it's a footgun the singleton shape invites — consider ref-counting connect/disconnect, or document that `useSocket` is single-instance only.

Also: the `// Depending on if you want the socket to stay alive across remounts … for now we'll match original behavior` comment reads like an unresolved TODO — worth turning into a decision or removing.

ui/src/hooks/useSocket.ts
```diff
@@ -1,274 +1,23 @@
...
-    triggerDiagnostics,
-    triggerStatus,
-    clearSession,
-    setClub,
-    simulateShot,
-    toggleDebug,
-    updateRadarConfig,
-    toggleCamera,
-    toggleCameraStream,
+    socketService,
```
jewbetcha: Minor: returning `socketService` here is redundant — `App.tsx` imports `socketService` directly from the module rather than pulling it off this hook's return value.

ui/src/services/socketService.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    this.socket.on('debug_reading', (data: DebugReading) => {
+      useDebugStore.getState().addDebugReading(data);
+    });
+
+    this.socket.on('radar_config', (data: RadarConfig) => {
+      useDebugStore.getState().setRadarConfig(data);
+    });
+
+    this.socket.on('camera_status', (data) => {
```
jewbetcha: Lost type annotation — `data` is implicitly `any` here (was `CameraStatus` before the refactor). socket.io's `.on` signature is `(...args: any[])`, so tsc won't flag it. Re-annotate `(data: CameraStatus)` to keep the payload typed.

ui/src/stores/useDebugStore.ts
```diff
@@ -0,0 +1,70 @@
+import { create } from 'zustand';
+import type { TriggerDiagnostic, TriggerStatus } from '../types/shot';
+import type { DebugReading, RadarConfig, DebugShotLog } from '../types/socket'; // Temporarily importing types from useSocket
```
jewbetcha: Stale comment — these types now come from `types/socket.ts`, not `useSocket`. Drop the `// Temporarily importing types from useSocket` trailer.

ui/src/stores/useShotStore.ts
```diff
@@ -0,0 +1,59 @@
...
+    },
+    clearShots: () => {
+      if (timerRef) clearTimeout(timerRef);
+      set({
+        latestShot: null,
+        shots: [],
+        isNewShot: false,
+      });
+    },
+    setIsNewShot: (isNewShot) => set({ isNewShot }),
```
jewbetcha: `setIsNewShot` appears to be unused — nothing in the diff consumes it. Drop it unless something off-diff depends on it.

clschnei: Ah, this is an internal state. Pretty much matched the previous provider methods 1:1. Will drop this.

ui/src/App.tsx
clschnei: 100% I skipped intentionally for the time being because my next refactor will be to co-locate selectors with the components that need them. Trying to keep the PR tightly scoped to just the new store. Currently all the state is prop-drilled down so the whole app is re-rendering regardless of if we use selectors. 

For the sake of idiomatic state management I will convert this because patterns matter :pray:

ui/src/hooks/useSocket.ts
clschnei: It might be worth not disconnecting at all honestly. If the application is tore down, that means the session closed.

clschnei:
Actively rebasing to pull in the new sim states.

jewbetcha (approved):
Thank you!

PR #139, opened 2026-06-22
feat(kld7): two-ray launch angle + speed-cosine correction, with CLI cleanup

johnpacino:
## What does this PR do?

Adds an opt-in **two-ray multipath** vertical launch-angle estimator for the K-LD7, plus the tiered-confidence, far-net handling, test tooling, and docs needed to use and trust it. It's selected with `--kld7-vertical-estimator two_ray`; the existing geometry/naive estimators are untouched and remain the default, so nothing changes unless it's explicitly turned on.

The pieces, all in service of that one estimator:

- **`two_ray.py`** — the estimator: per-frame two-ray demodulation that separates the ball from its floor reflection to recover true elevation; a tour-derived **Tier-1 / Tier-2 confidence classifier** (with a +boost for suppressed-reading shots); and **FSK range de-aliasing** for nets beyond the 16.4 ft range wrap.
- **`radc.py` / `tracker.py` / `server.py`** — wiring: estimator selection, the `--net-distance` arg, and a test-mode `--kld7-bypass-vertical-gate` flag (auto-enabled when two_ray is selected) that surfaces the radar angle for every shot the estimator catches.
- **`scripts/analysis/session_shot_report.py`** — an offline tool that replays a session through two_ray and renders a per-shot live-vs-offline HTML report (doubles as a regression lens).
- **`docs/kld7-launch-angle-explained.html`** — a visual explainer of the method, the tiers, and the confidence dots.

## Why was this required?

Indoors, the K-LD7's launch angle reads **systematically low**. The cause is ground multipath: the radar sees the ball *and* its reflection off the floor — a mirror-image "ball" below the surface — and the two blend, dragging the measured elevation down ("suppression"). The existing geometry/naive estimators average across that corruption rather than modeling it, so indoor launch angles came out several degrees low.

The two-ray estimator models the floor reflection explicitly and solves for the true ball elevation, then grades each shot into a confidence tier so the UI can distinguish a clean measurement from a corrected estimate. On the TrackMan-paired 6/15 7-iron session, clean (Tier-1) shots reach **0.69° MAE (P50 0.35°)**. It is **opt-in and default-off**, so it carries no risk to the shipping geometry path.

## Automated tests

Full suite: **897 passing**, pylint **9.69/10**, ruff clean.

- **`tests/test_kld7_two_ray.py`** (+497) — the estimator end-to-end on *physically exact* synthetic RADC payloads (the ball + floor-image phase, Doppler aliasing, the multipath fringe, and the F1B range all emerge from trajectory geometry rather than injected tones):
  - `two_ray_fit` component recovery; `estimate_two_ray` launch-angle recovery and refusal paths (DC blind zone, no impact timestamp, no frames, tau-implausible).
  - **De-aliasing**: default net is a no-op (byte-identical), the gate flips exactly at the unambiguous range, a far-net case unwraps wrapped far-flight frames, and the ceiling is the net slant.
  - **Tier classifier**: Tier-1 clean gate, Tier-2 boost vs no-boost confidences (0.35 / 0.65), and the per-club tour-derived configs (uniform derivation; 7-iron reproduced exactly).
- **`tests/test_session_shot_report.py`** (+185) — the report tool's session parsing, club/tuning extraction, no-data replay short-circuits, HTML rendering, and `main`.
- **`tests/test_server.py`** (+57) — `--kld7-bypass-vertical-gate` selection logic (default rejects an out-of-lane garbage read; bypass accepts any candidate as `gate_bypassed`) and `init_kld7` forwarding of `--net-distance`.
- **`tests/test_kld7.py`** — extract-call wiring for the net-distance parameter.

## Manual (human) testing

Verified on real hardware (Raspberry Pi + K-LD7 dual-radar + OPS243-A), across several indoor range sessions, with a TrackMan running alongside for ground truth:

- **Accuracy — 6/15, 7-iron, TrackMan-paired:** offline replay vs TrackMan gives **Tier-1 MAE 0.69° / P50 0.35°** (7 of 17 matched shots reached Tier-1); overall (incl. lower-confidence Tier-2) **1.23° MAE** on 17 shots. Confirmed the offline replay reproduces the *live* two_ray angles to ≤0.1°, so the offline numbers reflect what actually shipped.
- **De-aliasing:** byte-identical on real ≤10 ft-net sessions — `dealias=False` on every shot across 6/15 and three 6/18 sessions, MAE unchanged; a synthetic 25 ft-net case recovers 3 wrapped far-flight frames and stays under the net-slant ceiling.
- **Live two_ray — 6/20, 22× 7-iron on the Pi:** ran end-to-end, then analyzed with `session_shot_report.py` → 0 Tier-1 / 18 Tier-2 / 4 refused, consistent with the known indoor-suppression regime; spot-checked each non-rendered shot's reject reason against the live log.
- **Bypass flag:** with `--kld7-bypass-vertical-gate` (auto-on for two_ray), every two_ray candidate renders — including the implausible low reads the guards normally drop — as intended for test mode.
- **Mock mode:** `scripts/start-kiosk.sh --mock` boots and serves.

Edge cases exercised by hand: far-net de-aliasing, single-frame (Tier-2) shots, clutter-swamped shots that two_ray refuses (and correctly fall back to geometry), and the no-club path.

## Checklist

- [x] **Single feature/fix** — scoped to the two-ray estimator and its supporting tooling/docs
- [x] **Automated tests included** — new tests cover the estimator, tier classifier, de-aliasing, report tool, and server flags
- [x] **Manual testing described** — real-hardware + TrackMan validation documented above
- [x] Python tests pass (`uv run pytest tests/ -v`) — 897 passed
- [x] Pylint passes (`uv run pylint src/openflight/ --fail-under=9`) — 9.69/10
- [ ] UI builds — n/a, no `ui/src` changes in this PR
- [ ] UI lint passes — n/a, no UI changes
- [x] CHANGELOG entry — added under `[Unreleased] → Added`
- [x] No unrelated changes mixed in — three-dot diff is 11 files, all this feature

---

## Update — scope since first opened

This PR now covers the full K-LD7 launch-angle pipeline plus a CLI cleanup, not just the estimator:

- `--kld7` enables the two-ray vertical estimator (fixed cascade) **and** the ball-speed cosine correction by default.
- Guard-tripped radar reads show as **marginal (one-dot) confidence** instead of silently reverting to the club estimate.
- `--calculated-spin` (opt-in, off by default): kinematic spin from ball speed + launch angle; measured value kept in `spin_rpm_measured`.
- New `docs/kld7.md` (setup, measuring the rig, args, session reviewer).

## Breaking changes (flag cleanup)

- **Removed:** `--kld7-vertical-estimator` (estimator is now a fixed cascade), `--kld7-geometry` (kiosk preset), `--ball-speed-cosine-correction` (folded into `--kld7`).
- **Renamed:** `--kld7-bypass-vertical-gate` -> `--kld7-vertical-raw`; `--experimental-kld7-raw-radc-logging` -> `--kld7-raw-logging`.
- **Now required:** `--kld7-mount-tilt` must be supplied with `--kld7` (no safe default; measure with a phone inclinometer).
- Launch-angle source/confidence semantics changed.
- Coleman's `--experimental-kld7-*` tuning knobs are left untouched.

Full suite green (905 passed), pylint 9.69.


johnpacino:
@jewbetcha - branch updated. 

- **Removed:** `--kld7-vertical-estimator` (the estimator is now a fixed
  cascade: two_ray → geometry → single-frame geometry → naive),
  `--kld7-geometry` (kiosk preset), and `--ball-speed-cosine-correction`
  (folded into `--kld7`).
- **Renamed:** `--kld7-bypass-vertical-gate` → `--kld7-vertical-raw`;
  `--experimental-kld7-raw-radc-logging` → `--kld7-raw-logging`.
- **Now required:** `--kld7-mount-tilt` must be passed with `--kld7` (no safe
  default — a wrong tilt silently biases the launch angle; measure it with a
  phone inclinometer). `--kld7-angle-offset` defaults to the calibrated 1.5.

Net effect: `--kld7` now delivers the whole launch-angle pipeline by default,
and launch-angle source/confidence semantics change accordingly. Existing
command lines that used the removed/renamed flags will need updating —
`start-kiosk.sh` and `docs/kld7.md` are already updated to match.

I intentionally left the `--experimental-kld7-*` RADC tuning knobs untouched
since those are yours to clean up if/when you want. Also left some pre-existing
`ruff format` debt alone to keep this diff focused — happy to do that as a
separate PR if you'd prefer.

jewbetcha (approved):

PR #140, opened 2026-06-23
Update link and price for 27W USB-C Power Supply

mkneuffer:
Correcting link to Power supply on parts list.

Existing link goes to strain gauge. (Adafruit HX711 24-bit ADC for Load Cells / Strain Gauges)

New link goes to the correct 27W USB-C power supply, 5v 5A (Official Raspberry Pi 27W PD Power Supply 5.1V 5A with USB C)


jewbetcha (approved):
Thank you!

PR #157, opened 2026-07-24
feat: Swing speed training

Reillybags:
## Summary

Adds an experimental swing speed training mode for OPS243-based OpenFlight setups.

This mode is intended for club-only / air-swing training where there is no ball impact sound, such as driver air swings, SuperSpeed, The Stack System, Rypstick, and similar overspeed training aids.

## What changed?

- Added `--swing-speed` mode using OPS243 live speed reporting
- Added `--mock-swing-speed` mode for testing without radar hardware
- Emits swing speed reps into the existing UI shot/session flow
- Adds swing speed live display, stats, shot list support, delete support, CSV export/validation fields
- Adds training implement picker for Driver, SuperSpeed, The Stack, Rypstick, and Custom
- Adds player selection so multiple users can compare sessions
- Scopes live best/average/count by current player and selected training implement
- Adds radar tuning sliders for swing speed lower/upper gates
- Adds max speed guard for implausible high outliers
- Adds audio cue when a swing is captured
- Avoids saving `A!` during runtime radar mode restore

## Why was this required?

OpenFlight currently relies on rolling buffer mode plus a sound/impact trigger. That works for ball striking, but not for swing speed training where there is no impact sound.

This adds a dedicated training mode so OpenFlight can be used for overspeed work and air swings while keeping normal launch monitor mode as the default.

## Automated tests

Focused tests passing locally:

```text
19 passed, 1 warning

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,9 @@ scripts/start-kiosk.sh --iwr6843 \
 # With K-LD7 launch-angle geometry defaults (deprecated hardware)
 scripts/start-kiosk.sh --kld7-geometry
 
+# Swing speed training (air swings / speed sticks, no sound trigger)
+scripts/start-kiosk.sh --swing-speed
+
 # Development mode (no hardware)
 scripts/start-kiosk.sh --mock
 ```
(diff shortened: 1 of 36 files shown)


jewbetcha:
I'm testing this now, is there a reason for renaming the rolling buffer API commands? This will break the existing functionality

jewbetcha (approved):

PR #159, opened 2026-07-25
feat: add experimental multitaper spin estimator

johnpacino:
## What does this PR do?

Adds the best-performing offline OPS243-A spin track to the live rolling-buffer
pipeline as an explicitly experimental estimator.

The estimator runs automatically for every valid rolling-buffer capture. It
reports its best bounded spin candidate without confidence, club, or rail gates
so Coleman and other testers can collect complete evaluation datasets instead
of losing difficult shots to silent rejection.

Reported shots now include:

- `spin_rpm`
- `spin_method: "multitaper_ungated"`
- `spin_quality: "experimental"`
- peak-to-floor evidence through the existing spin confidence/SNR fields
- the fitted `spin_multipath_fade_hz`

The normal and full-screen UIs display **EXPERIMENTAL** below Spin Rate. These
readings intentionally do not drive spin-adjusted carry.

### How the estimator works

1. Repairs clipped 12-bit I/Q samples by interpolating across samples near the
   ADC rails, then removes the I/Q DC offsets.
2. Bandpasses the complex signal around the measured ball Doppler frequency and
   extracts its amplitude envelope. The window starts at ball detection,
   removes filter transients, and ends at sustained ball-signal loss when the
   net impact truncates the return.
3. Estimates the dominant 25-90 Hz two-ray ground-fade component. A least-squares
   model removes polynomial drift, that fade frequency, and its second harmonic.
4. Applies four DPSS (Slepian) tapers with time-bandwidth 2.5 to the residual
   envelope. It computes a spectrum for each taper and averages their power.
5. Selects the strongest coherent peak in the 1,500-11,000 RPM band. The
   multitaper peak-to-median-floor ratio is retained as evidence, but is not
   currently a reporting gate.

This branch is based on current upstream `main` and has no dependency on the
IWR6843 work.

## Why was this required?

OmniPreSense recommended overlapping 128-sample speed windows to expose the
small ripple associated with ball rotation. Our TrackMan-paired analysis found
that the strongest ripple was usually not spin: it tracked ball speed and radar
placement, consistent with two-ray ground-bounce fading. The 1x seam tone was
present, but could be 10-30x weaker than that placement-dependent fade.

A single-window periodogram is sensitive to the exact endpoints of this short,
non-stationary capture. Spectral leakage and high variance can move the selected
peak between neighboring or unrelated components. DPSS tapers are designed to
concentrate energy inside a controlled bandwidth. Averaging several tapered
spectra reduces variance and endpoint leakage without requiring a longer radar
capture.

Multitaper was selected because it performed best among the offline tracks we
evaluated against TrackMan, not because it has reached production accuracy.
Blind evaluation remains above the product target:

| Session | MAE | Median absolute error |
| --- | ---: | ---: |
| July 14 | 2,499 RPM | 1,682 RPM |
| July 17 held-out | 3,073 RPM | 2,968 RPM |

This PR is intended to broaden hardware evaluation and collect enough paired
data to improve the multipath model. A reported candidate is not a claim that
the reading is correct. The estimator can still choose residual fade energy,
harmonics, or an RPM search rail. Captures with invalid processing or
insufficient post-impact ball signal can still return no spin.

### Reviewer focus

- Is the fitted fade plus second-harmonic model appropriate across expected
  indoor geometries?
- Is the 25-90 Hz fade search band broad enough without removing real low-spin
  energy?
- Is automatic, ungated reporting appropriate for an evaluation release?
- Does Raspberry Pi processing latency remain acceptable during live use?
- Are the `experimental` UI and carry-isolation semantics clear enough to avoid
  treating candidates as production measurements?

## Automated tests

- `uv run pytest tests/ -q`: 919 passed, 6 skipped
- `uv run pytest tests/test_multitaper_spin.py -q`: 6 passed
- `uv run pylint src/openflight/ --fail-under=9`: 9.70/10
- `uv run ruff check src/openflight/ tests/test_multitaper_spin.py`
- `cd ui && npm run test`: 20 passed
- `cd ui && npm run build`
- `cd ui && npm run lint`
- `cd ui && npm run format:check`

The automated tests cover clipped-I/Q repair, synthetic seam recovery beneath a
stronger fade tone, automatic processor selection, ungated reporting, carry
isolation, API serialization, and the UI's experimental label.

## Manual (human) testing

The stored TrackMan-paired captures were replayed through the offline-selected
multitaper implementation. The live implementation reproduced all 193 stored
offline predictions exactly, including the July 14 and held-out July 17 sets
reported above.

Live Raspberry Pi/OPS hardware latency has not yet been benchmarked on this
branch. That is an explicit goal of this evaluation PR rather than a completed
manual test.

Suggested evaluation for Coleman and other testers:

1. Update to this PR and start OpenFlight normally with
   `scripts/start-kiosk.sh`; there is no estimator flag to enable.
2. Confirm the OPS persistent rolling-buffer setup is already complete.
3. Record radar height, ball distance, tilt, net distance, club, and ball model
   for each session. Geometry matters because the dominant fade shifts with
   placement.
4. Collect TrackMan-paired shots and retain the OpenFlight JSONL session logs.
5. Compare every candidate, including low-evidence and search-rail results,
   rather than evaluating only successful-looking shots.

## Checklist

- [x] **Single feature/fix** - this PR is scoped to one experimental estimator
- [x] **Automated tests included** - backend and UI behavior are covered
- [x] **Manual testing described** - offline replay and hardware gap documented

### DIFF
--- a/docs/rolling_buffer_spin_detection.md
+++ b/docs/rolling_buffer_spin_detection.md
@@ -10,6 +10,28 @@ The OPS243-A radar can detect golf ball spin by analyzing micro-variations in th
 2. **Overlapping FFT Windows** - Increases temporal resolution to see spin "waviness"
 3. **Secondary FFT** - Extracts spin frequency from speed oscillations
 
+## Experimental Live Estimator
+
+The live rolling-buffer pipeline automatically runs the ungated multitaper
+estimator for every valid capture. No estimator flag or additional setup is
+required beyond the normal persistent rolling-buffer configuration.
+
+- The UI displays the candidate below Spin Rate with an **EXPERIMENTAL** label.
+- The result is logged with `spin_method: "multitaper_ungated"` and the detected
+  multipath fade frequency for later TrackMan comparison.
+- The estimator intentionally reports its best 1,500-11,000 RPM candidate
+  without a confidence, rail, or club-selection gate so evaluation sessions do
+  not silently discard difficult shots.
+- Experimental spin does not currently drive spin-adjusted carry. A candidate
+  can be useful for model evaluation without being reliable enough for flight
+  calculations.
+
+TrackMan-blind validation remains above the production target: July 14 had
+2,499 RPM MAE (1,682 RPM median absolute error), and the held-out July 17
+session had 3,073 RPM MAE (2,968 RPM median). Treat individual readings as
+evaluation data rather than ground truth until geometry and multipath rejection
+improve.
+
 ## Current vs Rolling Buffer Approach
 
 | Aspect | Current (Streaming) | Rolling Buffer |
--- a/src/openflight/launch_monitor.py
+++ b/src/openflight/launch_monitor.py
@@ -223,7 +223,9 @@ class Shot:
         launch_angle_horizontal_source: Source for horizontal launch angle
         spin_rpm: Spin rate in RPM (from rolling buffer mode)
         spin_confidence: Confidence in spin measurement (0-1)
+        spin_method: Estimator that produced the spin candidate
         spin_result_quality: Processor quality label for the spin detection
+        spin_multipath_fade_hz: Fitted two-ray fade frequency, when available
         spin_snr: Signal-to-noise ratio of the spin envelope peak
         spin_modulation_depth: Envelope std/mean inside the spin window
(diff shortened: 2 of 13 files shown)


jewbetcha (commented):
👌👌

jewbetcha (approved):

PR #162, opened 2026-07-27
Change Raspberry Pi 5 price from $60 to $130

Elkhunder:
Updated the price of the Raspberry Pi 5 in the parts list.


jewbetcha (approved):
Thanks, and also sad 😭

PR #168, opened 2026-07-30
feat(iwr6843): include temperature report in dumps

johnpacino:
## What does this PR do?

Adds IWR6843 RF temperature reporting to the production L3 dump path.

The firmware now appends a v5 24-byte temperature report after the fixed dump header when `rlRfGetTemperatureReport(...)` succeeds. The host parser, dump sizing logic, capture monitor, server integration, and session logger now understand that optional v5 header extension and record it on `iwr6843_capture` JSONL entries.

This keeps the existing windowed range-snapshot geometry intact while adding lightweight thermal telemetry.

## Why was this required?

The IWR6843 is now used as the angle radar, and the board can feel hot during normal operation. Without logging internal temperature, we only had subjective "hot to touch" checks and no way to correlate thermal behavior with sessions, enclosures, airflow, or capture failures.

The first attempt to roll this out also exposed a host-side fixed-header assumption: v5 dumps are 24 bytes larger, so dump sizing and frame-window parsing must use the parsed `header_nbytes` instead of assuming the original 20-byte header.

Follow-up hardening from PR feedback:

- Reject future unsupported dump versions loudly.
- Preserve the previous `_v2.bin` release artifact for rollback.
- Add a dated temperature release image: `l3_dump_vTX2_hwa_window53_12loops_18frames_4ms_temperature_report_20260731.bin`.
- Document that TI mmWaveLink temperature fields are signed with `1 LSB = 1 deg C`.
- Add malformed-dump tests for short temperature extension, short window metadata, truncated payload, and v5 read sizing.

## Automated tests

Added or updated tests for:

- v5 temperature report pack/parse round-trip
- v5 windowed snapshot frame-window alignment
- TX-pair projection preserving the temperature report
- capture monitor retaining `temperature_report` on `IWR6843Capture`
- session logger writing `temperature_report` and defaulting it to `null` when absent
- future-version rejection
- malformed v5/windowed/truncated dump rejection
- serial `read_dump` sizing for v5 header extensions
- production release filename

Commands run:

```bash
uv run pytest tests/test_iwr6843_pipeline.py tests/test_iwr6843_monitor.py tests/test_session_logger.py tests/test_iwr6843_driver.py tests/test_iwr6843_firmware_rearm.py -q
```

Result: `86 passed, 1 warning`

```bash
uv run ruff check src/openflight/iwr6843/dump.py src/openflight/iwr6843/driver.py src/openflight/iwr6843/monitor.py src/openflight/server.py src/openflight/session_logger.py tests/test_iwr6843_pipeline.py tests/test_iwr6843_monitor.py tests/test_session_logger.py tests/test_iwr6843_driver.py tests/test_iwr6843_firmware_rearm.py
```

Result: `All checks passed!`

## Manual (human) testing

Built the firmware from the clean branch in the TI SDK container:

```bash
docker run --rm -v /Users/john.pacino/Projects/openflight-temperature-report:/work -w /work/firmware openflight-iwr-sdk:latest make build-native
```

Current release binary:

```text
firmware/releases/l3_dump_vTX2_hwa_window53_12loops_18frames_4ms_temperature_report_20260731.bin
size: 339,780 bytes
sha256: 8a87593954fd5ae2b7adf709c78626f81a87b8988b8dc28f2b2be7b5c99eac12
```

Rollback binary preserved:

```text
firmware/releases/l3_dump_vTX2_hwa_window53_12loops_18frames_4ms_v2.bin
sha256: 3045bb2f087b40c228bf1dd5190cf3fac6dbde50682c7927e86714314b0e7fcb
```

Flashed and ran on the Raspberry Pi using:

```bash
./run-openflight.sh
```

Manual clap test observed:

```text
[IWR6843] Capture #1 complete: 549566 bytes in 5.33s
```

Confirmed the session JSONL contains the new temperature report:

```json
"temperature_report": {
  "device_time_ms": 2910634,
  "rx0_c": 48,
  "rx1_c": 48,
  "rx2_c": 48,
  "rx3_c": 49,
  "tx0_c": 49,
  "tx1_c": 49,
  "tx2_c": 49,
  "pm_c": 52,
  "dig0_c": 52,
  "dig1_c": 53
}
```

The clap was expected to withhold radar launch angle as `rejected_by_ball_tracker`; it verified trigger, dump transfer, v5 parsing, and session logging.

## Checklist

- [ ] UI builds (`cd ui && npm run build`)
- [ ] UI lint passes (`cd ui && npm run lint`)


dieterolson:
Nice clean plumb-through — the `pack_dump` both-direction invariant and the firmware's v4 fallback on a failed temp read are great.

One hardening suggestion while the version field is getting its first real use: `parse_header` accepts any version, and the failure mode for a mismatched pairing is silent. Concretely, an older host reading a v5 dump from this firmware: `payload_nbytes` is version-blind, so `read_dump` sizes the read 24 bytes short and truncates; `_validate_dump` uses the same formula, so validation passes; `parse_dump` (18-frame windowed format) then reads the first 18 bytes of the temperature struct as `range_bin_starts` and the int16 body lands 24 bytes shifted — no exception anywhere, just a garbage cube. Likely user-visible symptom: every shot rejecting (e.g. `rejected_by_ball_tracker`) with nothing pointing at the firmware/host mismatch. (Mechanism traced in code; not reproduced on hardware.)

The `if ver >= 5:` gate also assumes every future version carries exactly this 24-byte extension, so a v6 with a different layout would misparse the same way on today's parser.

Suggestion: a ceiling in `parse_header` — `if ver > L3_DUMP_VERSION_TEMPERATURE: raise ValueError(...)`. It can't help hosts already in the field, but it makes the *next* version bump fail loudly instead of silently. Happy to follow up with a PR adding the guard plus dump-parsing edge-case tests (truncated payload, short window table, short temp extension) once this lands.

dieterolson:
Question on release management: this updates `firmware/releases/..._v2.bin` in place, so the validated v2 build leaves the tree and the `_v2` name no longer matches the content (now emitting header v5). Two knock-ons: rollback to the last field-validated image becomes git archaeology, and `feat/club-path-iwr`'s 25-frame image was built pre-this-PR (emits v4, no temperature block) — once both land, flashing the newer firmware silently drops the telemetry this PR adds, and "current release" is ambiguous. Would you take a new filename per build (keeping the old binary) and a short note separating the two version axes — filename "vN" (build iteration) vs header version (wire format)? Happy to do the docs part.

dieterolson:
Two small ones: (1) the session-log keys are `rx0_c` etc., implying °C, but the `rl*Sens` int16s are logged unscaled — is the DFP already returning °C? Worth a one-line comment citing the unit either way so nobody plots degrees off raw counts. (2) The new error paths are untested (`short temperature report extension`, both `pack_dump` ValueErrors, `read_dump` sizing on a v5 header) — happy to add those in a follow-up test PR rather than grow this one.

johnpacino:
Thanks, agreed on making future wire-format mismatches fail loudly. Added this in `55fd526`:

- `MAX_SUPPORTED_DUMP_VERSION = 5`
- `parse_header(...)` now raises `ValueError` for dump versions above that ceiling.
- Added regression coverage for future-version rejection, short temperature extension, truncated payload, short window metadata, and v5 serial `read_dump` sizing.

That does not help already-deployed older hosts parse v5, as you noted, but it should prevent the next version bump from silently shifting payload bytes.

johnpacino:
Yep, the TI mmWaveLink docs say the `rlRfTempData_t` temperature fields are signed with `1 LSB = 1 deg C`, so the `_c` session keys are correct.

I added comments in `55fd526` on both sides of the wire contract:

- `firmware/iwr6843/dump_format.h`
- `src/openflight/iwr6843/dump.py`

I also added the missing error-path coverage in the same commit:

- short temperature report extension
- both `pack_dump` temperature/version mismatch errors
- v5 `read_dump` sizing
- future dump version rejection
- short per-frame window table
- truncated payload

Focused validation is now `86 passed, 1 warning`, and targeted Ruff is clean.

johnpacino:
Good call. I changed the release handling in `55fd526`:

- Restored `firmware/releases/l3_dump_vTX2_hwa_window53_12loops_18frames_4ms_v2.bin` to the original rollback artifact.
- Added the current image as `firmware/releases/l3_dump_vTX2_hwa_window53_12loops_18frames_4ms_temperature_report_20260731.bin`.
- Updated `firmware/Makefile` and the docs to point at the dated temperature-report image.
- Added docs explaining that the release filename/build identity is separate from the dump header/wire-format version.

Hashes are now:

```text
temperature_report_20260731: 8a87593954fd5ae2b7adf709c78626f81a87b8988b8dc28f2b2be7b5c99eac12
rollback _v2:                 3045bb2f087b40c228bf1dd5190cf3fac6dbde50682c7927e86714314b0e7fcb
```

jewbetcha (approved):

PR #217, opened 2026-08-18
feat(mobile): add React Native (Expo) companion app

dchasepdx:
## Summary

Adds a React Native (Expo, SDK 54) mobile companion app under `mobile/`. It connects to the existing Flask/Socket.IO server over the LAN and renders live shots.

## Why was this required?

The README lists a phone app as a wanted contribution. A native app is the foundation for future device-only capabilities (BLE/watch, GPS-based features) a browser can't reach. This PR is the scaffold plus a first useful screen; it does not touch the Python server or the web UI.

## Automated tests

None in this PR. `mobile/` has no test harness yet, and the repo's "tests included" PR gate only checks `src/openflight/**` and `ui/src/**`, so `mobile/` changes don't trigger it. Standing up jest-expo + React Native Testing Library is planned as a separate follow-up.

## Manual (human) testing

- `npx tsc --noEmit` clean; `npx expo export --platform web` bundles clean.
- Connected to a local `--mock` server, used Simulate Shot, and verified the current-shot tiles render — including the conditional Club Path / Spin Axis tiles and the confidence dots.


jewbetcha:
Thanks for this! Would love for the community to test this out and provide some feedback before merging

dchasepdx:
> Thanks for this! Would love for the community to test this out and provide some feedback before merging

Yeah no problem! Let me add some instructions for running the dev build connected to the mock server

mobile/App.tsx
```diff
@@ -0,0 +1,230 @@
...
+  const latestShot = shots[0] ?? null;
+  const isConnected = connectionState === 'connected';
+
+  const disconnect = useCallback(() => {
+    socketRef.current?.close();
+    socketRef.current = null;
+    setConnectionState('disconnected');
+  }, []);
+
+  const connect = useCallback(() => {
```
Cormac131, line 47: When a connection attempt has failed you are not able to connect again. App has to be reloaded to allow a new connection after an error state

dchasepdx: That's a good call. I have another branch with more features that this bug is fixed on. I wanted to get a quick PR up to discuss whether or not React native was the right path. Here's the phase 0 code if you're curious: https://github.com/dchasepdx/openflight/pull/2

dchasepdx:
Closing this — superseded by the repo split.

The mobile app now lives in its own repo, [open-flight/openflight-mobile](https://github.com/open-flight/openflight-mobile), so this no longer targets the right place.

The work here isn't lost: I moved it with `git subtree split --prefix=mobile` and rebased onto the mobile repo's initial commit, so the original commits and authorship are preserved rather than squashed. It's now open as **[open-flight/openflight-mobile#1](https://github.com/open-flight/openflight-mobile/pull/1)**, together with the Phase 0 foundation work from #224 (navigation shell, socket service, session store, and a jest-expo harness with 20 passing tests).

The `Makefile` `mobile-dev` target from this PR is intentionally not carried over — it has no meaning outside the monorepo, and nothing mobile-related ever landed on `main` here, so no cleanup is needed on this side.

PR #234, opened 2026-08-22
docs(parts): add CAD pricing column and refresh stale USD prices

noahjobse:
While pricing out an actual build from Canada I ended up re-checking every line in `docs/PARTS.md` against its vendor. A few numbers had drifted and two of the Amazon links are misleading outside the US, so this PR fixes the pricing and adds Canadian sourcing.

### Price corrections (captured 2026-08-22 from vendor sites)

| Item | Doc said | Vendor says |
|---|---|---|
| OPS243-A | $249 | **$224** (and currently *"Available on backorder"*) |
| IWR6843LEVM | $150 | **$156.45** at TI |

The **Cost Summary did not reconcile with its own tables**. Core was listed at $355 while the three core rows summed to $425 at the old prices, and Power & Accessories at $27 against $29. Corrected USD total is **~$609**, not ~$556. Every category row now sums to its line items (I checked this programmatically).

### Canadian sourcing

Added a `~CAD` column to each table plus a **Buying in Canada** section. Most of the build is stocked domestically:

- **PiShop.ca** for the Pi 5 ($153.95), 27W PSU ($16.95), microSD ($28.95)
- **DigiKey Canada** for the SEN-14262 ($20.76, free shipping over $100 CAD)
- **Elmwood Electronics** for the LIS3DH ($7.99) and the STEMMA QT cable ($1.99, versus the ~$10 USD Amazon kit)

Two parts have no Canadian distributor and must be imported. Worth flagging: **order the LEVM from TI direct**, since DigiKey Canada lists the same board at **$297.83 CAD** against roughly $215 CAD landed from TI.

### Two link gotchas

Amazon `.com` ASINs do not carry over to Amazon.ca, and both of the ones I checked resolve to a *different product* there:

- The display link lands on an 800x480 MIPI DSI panel rather than a 1024x600 unit.
- The Layout B hub link lands on a **bus-powered** hub. That one is worth a warning on its own merits: Layout B exists precisely because the Pi cannot supply both radars, so a bus-powered hub fails to solve the problem it is there for. I added a caution note.

### Notes

- CAD values are Canadian distributor list prices where one stocks the part, otherwise the USD price converted at the Bank of Canada daily rate of **1.3760 (2026-08-21)**. Converted figures are marked `~` and exclude shipping, duty, brokerage and tax.
- The deprecated K-LD7 table is deliberately left USD-only, since those parts should not be bought for new builds.
- Docs-only change; no code or tests touched.

Happy to drop the CAD column and keep just the price corrections if you would rather not carry regional pricing in the main BOM.

### DIFF
--- a/docs/PARTS.md
+++ b/docs/PARTS.md
@@ -6,13 +6,19 @@ Hardware components for building the OpenFlight golf launch monitor.
 
 > **Next step after gathering parts:** See the [Raspberry Pi Setup Guide](raspberry-pi-setup.md) for assembly and software installation.
 
+> **Prices:** USD figures are the vendor's own list price. CAD figures are either a
+> Canadian distributor's list price (where one stocks the part) or the USD price
+> converted at the Bank of Canada daily rate of **1.3760 USD/CAD (2026-08-21)**.
+> Converted figures are marked `~` and **exclude shipping, duty, customs brokerage,
+> GST/HST/PST**. See [Buying in Canada](#buying-in-canada) before ordering.
+
 ## Core Components
 
-| Part | Description | Link | ~Price |
-|------|-------------|------|--------|
-| **OPS243 Radar** | Doppler radar for ball/club speed detection | [OmniPreSense](https://omnipresense.com/product/ops243-doppler-radar-sensor/) | $249 |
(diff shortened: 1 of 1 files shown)


noahjobse:
Closing: opened against upstream by mistake, this was meant for my own fork. Apologies for the noise.

PR #223, opened 2026-08-19
feat(kiosk): Redesign kiosk UI as an instrument panel with a 10-metric live table

Cormac131:
## What does this PR do?
Screenshot 2026-08-20 180739
Screenshot 2026-08-20 180729
Screenshot 2026-08-20 180819
Screenshot 2026-08-20 180807
Screenshot 2026-08-20 180750

Replaces the old top-header / full-screen interstitial kiosk with a panel shell: bottom footer, Live / Stats / Shots / Camera / Players / Debug tabs, and a new tree under `ui/src/components/panel/`.

On Live, every shot is a **fixed 10-metric table** (including Club AoA). All ten values share one fitted font size so a wide number does not shrink one tile. Tapping a tile pins it top-left (accent title) and, after a new shot, that metric **spotlights full-screen for 10s** (tap to dismiss). Club name sits in the panel header; change club / training implement is a full-screen picker. The header LED opens server / radar / ball-detection status. Theme, units, language (EN / ES / FR / PT), simulators, and shutdown live in a slide-up menu behind the footer logo. Battery and charging sit in the footer. Stats club chips and the shots list drag-scroll on the kiosk. Display mode, Launch Daddy, and remaining views share the same metric cards and dark/light tokens.

Old one-off screens (`ShotDisplay`, `StatsView`, `ShotList`, `CameraFeed`, `ClubPicker`, `ClubSelectScreen`, `PlayerPicker`, etc.) are gone.

## Why was this required?

The old live view made metrics hard to read on the kiosk. Values were small, the layout reflowed, and it was easy to miss the number that mattered after a shot.

This change is to **improve usability and make viewing metrics clearer**: a fixed 10-metric table so every number stays visible, a pinned/spotlighted selected metric so the important value is large after a shot, and chrome that fits the 7-inch Pi without covering the grid. Players, language, units, theme, and battery are reachable without burying them in the live layout. Dark/light tokens and shared primitives keep Display Mode and Live in the same visual system. Without this, shot review stays cramped and hard to scan.

## Automated tests

New/updated Vitest coverage:

- **Live table / spotlight:** `LivePanel.test.tsx`, `liveMetrics.test.ts` — 10-tile grid, pin-to-top-left, 10s spotlight on new shots (not restored), shared fitted font size, swing-speed 5-tile row, ball-detection warning, estimated marks
[...]

PR #239, opened 2026-08-26
feat(kiosk): add first-run kiosk setup and persisted live view modes

Cormac131:
## What does this PR do?

Adds a **first-run kiosk setup** so a new OpenFlight does not dump the player straight onto Live.

On first launch (no `openflight.onboarding.completed:v1`), the shell is replaced by a fullscreen wizard: language + units, theme, then Live view (tiles / timed preview / hold preview). Completing it writes that flag, selects Driver, and lands on Live **without** opening the club picker that one time. Later launches still open the picker. `/display` never shows the wizard.

Choices persist and stay editable in the footer menu:

- language: `openflight.locale:v1`
- units: `openflight.unit-system`
- theme: `openflight.theme`
- live view: `openflight.live-view:v1` (`tiles` | `timed` | `sticky`, duration 5/10/15s)

Live can show the selected metric full-screen after a new shot (timed: hide after the chosen duration; hold: until tap). Shut down was removed from the menu sheet; use the **header** power button. Ball detection / simulators stay in the header status menu, not the sheet.

Menu live-view segments are short (**Tiles / Timed / Hold**) so three options fit; labels ellipsis if a translation is longer.

## Screens
**Welcome** 
image

**Language & units**
image

**Theme**
image

**Live view**
image

**Finish**
image

**Updated menu**
image
image

## Why was this required?

A first boot currently assumes language, units, theme, and Live layout are already known. On a Pi kiosk that is a bad first minute: the club picker appears before any of those choices, and live preview modes were not persisted. Without this, every new unit needs a hidden menu tour, and timed/hold overlays either do not exist or cannot be changed later.

## Automated tests

- **Stores:** `useOnboardingStore` (missing / present / write / quota failure / no `window`); `useLiveViewStore` (persist, invalid JSON fallback, duration).
- **Wizard:** `OnboardingFlow.test.tsx` — welcome, language+units, theme tiles, duration chips only on Timed (inside that tile), Start on done.
- **App gate:** `App.test.tsx` — wizard vs shell based on the completed flag.
- **Live overlay:** `useShotSpotlight.test.ts` plus Live panel coverage for timed/hold vs tiles.
- **Menu:** `MenuSheet.test.tsx` — Tiles/Timed/Hold, duration only when timed, no System / ball detection / simulators, no Shut down in the sheet. `PanelHeader.test.tsx` — header still opens shutdown. `SegmentedControl` — pressed state + label wrapper used for ellipsis.
- **i18n:** catalog lockstep for EN/ES/FR/PT (`menu.liveTiles` / `liveTimed` / `liveHold` plus onboarding strings).
- **Playwright** `ui/tests/e2e/onboarding.spec.ts`: first visit is welcome not picker; full walk lands on Live + Driver; later launch opens picker; `/display` skips wizard; menu has live view and no sheet shutdown (header still shuts down); timed overlay hides after 5s, hold stays until tap. `gotoApp` defaults `onboarded: true` so existing e2e keep skipping the wizard.

## Manual (human) testing

**Vite (dev machine).** Ran the UI at `http://127.0.0.1:5173` at **800×480**. Seeded `openflight.onboarding.completed:v1` so the kiosk shell loaded, closed Select club, opened the footer menu.

Observed:

- Menu sections: Units, Theme, Language, Live view only — **no System**, **no Shut down** in the sheet.
- Live view segments: **Tiles / Timed / Hold** (English); Portuguese **Mosaicos / Tempo / Fixa**.
- Header still has Shut down.
- Injected a long segment label: it clipped with an ellipsis instead of overflowing the control.
- 
**Raspberry Pi kiosk.** Walked the same first-run and menu flows on a Pi (touchscreen / `scripts/start-kiosk.sh`), not only in the browser on a laptop.

## Checklist

- [ ] **Single feature/fix** — intended story is first-run setup + live view
- [x] **Automated tests included** — Vitest + Playwright as above
- [x] **Manual testing described** — Vite kiosk menu at 800×480, EN/PT, ellipsis
- [ ] No unrelated changes mixed in


Cormac131:
@jewbetcha This is to address the comments from the last pr: https://github.com/jewbetcha/openflight/pull/223#issuecomment-5414420322

johnpacino:
@jewbetcha - need your final review on this one.

johnpacino (changes requested):
@Cormac131 - This PR exceeds the 2000 line change limit and looks like it has 3 distinct features. Thoughts about breaking this up into 3 PRs? (cc: @jewbetcha )

**1. feat(kiosk): add persisted live view modes**

This should contain Tiles/Timed/Hold, persisted duration, menu controls, spotlight behavior, and their tests. Please also address:
[P1] Preserve experimental provenance in Live. buildLiveMetrics() sets experimental, but LivePanel neither passes it to MetricCard nor displays it in the spotlight.
[P2] Add vertical drag scrolling to the menu. At 800×400, selecting Timed puts the duration buttons below the visible menu, and Pi-style mouse dragging does not scroll it. Add mouse-drag and touch-tap E2E coverage.

**2. feat(kiosk): add first-run setup**
Build onboarding on top of the merged live-view modes. Language, units, theme, and live-view selection belong here. The shutdown control must remain available throughout onboarding, including the welcome screen, per the original “available at all times” requirement.

**3. refactor(ui): standardize experimental metric provenance**
Move the broader flask-icon/provenance redesign across MetricCard, DisplayMode, and liveMetrics into its own PR.

Also,  each PR should also include the required AI-assistance disclosure and completed checklist.

Cormac131:
@johnpacino - Thanks for the feedback. I’ll fix both concrete issues: P1 was an omission when I replaced the existing stretching provenance text with the compact experimental marker, rather than a new provenance feature. I’ll pass it through LivePanel and display it in the spotlight. I’ll also add drag scrolling and E2E coverage for the menu at 800×400.

I’m not convinced three PRs would improve reviewability. Live-view modes directly implement the feedback from the previous UI PR, and onboarding configures that same persisted state. The provenance change only fixes the existing text-stretching issue. Most of the 2,000+ lines are supporting code: 520 lines of CSS, roughly 600 lines of tests, plus repeated translations and documentation.

I interpreted shutdown being “available at all times” as applying to the normal kiosk shell, where it is now always in the header rather than behind the menu. Adding it to the pre-shell wizard introduces undefined resume/persistence behaviour if shutdown occurs midway through setup.

The checklist is already included and I’ve only ticked checks I actually ran. The AI policy was added after this PR was opened, but I’m happy to add an accurate disclosure. My preference is to fix P1/P2 and keep the connected implementation together.

johnpacino (approved):
@jewbetcha - i'm good with this being a single PR if you are. Just pending your approval and merge.

PR #240, opened 2026-08-26
perf: reduce shot-to-metrics latency

jewbetcha:
## What does this PR do?

- Publishes OPS243 metrics immediately, then performs IWR6843 and camera enrichment on one ordered background worker.
- Keeps the OPS capture callback non-blocking under queue and coordinator pressure: callbacks only register/queue work, while a dedicated finalization thread exclusively handles fallback calculation, persistence, WebSocket publication, and simulator forwarding.
- Bounds ordered finalization with a 20-second end-to-end watchdog and a four-shot coordinator cap; overdue work finalizes OPS-only and late hardware results are discarded.
- Immediately clears IWR/camera progress with an explicit skipped update when optional work cannot be admitted.
- Assigns a stable shot identity in `RollingBufferMonitor` before the raw OPS capture is logged, and reuses it for IWR, camera, final session, WebSocket, and simulator records.
- Keeps non-blocking IWR and camera progress visible until the matching final or skipped update arrives.
- Updates enriched shots in place without incrementing the new-shot version, preventing duplicate flashes, remounts, audio, or Launch Daddy effects.
- Reuses overlapping FFT windows, caches immutable multitaper inputs, reduces serial-buffer copying, and vectorizes circular phase-median scoring.
- Corrects the standard 128-sample FFT cadence documentation from approximately 56 Hz to approximately 234 Hz.

## Why was this required?

Users were waiting roughly five to seven seconds after impact before any metrics appeared because the UI emit was behind the IWR UART dump and optional camera processing. Earlier revisions made that work asynchronous but could block OPS rearming under queue/coordinator pressure, publish final shots and hardware artifacts out of order, retain later shots indefinitely behind a hardware call that never returned, or leave a skipped shot displaying stale hardware-progress feedback. This revision preserves the immediate OPS result, keeps the production capture callback responsive, maintains one identity and publication order across every record, and bounds both the hardware queue and finalization coordinator.

The startup-only camera auto-exposure work is in #244. The live-metric subtitle layout fix is in #246. Both are independently reviewable and releasable.

## Measured results

- OPS metrics no longer wait for the roughly seven-second IWR UART dump.
- Warm IWR host processing improved from 168.1 ms to 147.6 ms median, a 12.2 percent reduction.
- Circular phase median improved from 30.5 microseconds to 12.4 microseconds per call, about 2.5 times faster.
- A capacity-pressure regression blocks required finalization and proves the fifth OPS callback still returns, leaving trigger rearming independent of persistence, socket, and simulator latency.
- A 20-shot stuck-hardware regression keeps all coordinator collections at four shots or fewer while finalizing every valid OPS shot in order.
- Initial UI latency and hardware-enrichment latency are recorded separately in session pipeline timing.

## Automated tests

- Added concurrent coverage for non-blocking queue overflow and coordinator eviction, ordered final publication, simulator order, finalizer failure, and matching stable identities across raw OPS, IWR, camera, WebSocket, and session records.
- Added watchdog coverage proving a never-returning head finalizes OPS-only at its deadline, later overflow stays bounded, and late enrichment neither republishes nor mutates the finalized shot.
- Added UI coverage proving explicit skipped updates clear hardware-progress feedback without duplicating the shot.
- Full server test module: 183 passed, 1 skipped.
- Full UI unit suite: 245 passed.
- Production UI build, ESLint, Ruff, and Ruff formatting checks pass.
- Focused Pylint score: 9.70/10.

## Manual (human) testing

- Reviewed the socket sequence to confirm an immediate shot is followed by a timestamp-matched final `shot_update`, while queue/worker admission failures first publish an explicit skipped update that closes the pending indicator.
- Real OPS/IWR/camera hardware was not available locally, so final end-to-end device validation remains pending.

## Checklist

- [x] Single feature/fix
- [x] Automated tests included
- [x] Manual testing described
- [x] UI builds and lint/format checks pass
- [x] Ruff passes on modified Python files
- [x] Independent camera-exposure and subtitle work split out

### DIFF
--- a/docs/camera/README.md
+++ b/docs/camera/README.md
@@ -238,16 +238,10 @@ inside the configured frame period.
 At startup, the controller begins with the last good setting saved for the same
 resolution and frame rate. If that setting is not usable, it can jump several
 steps and recheck after `300 ms`, rather than waiting through one five-second
-step at a time. It makes at most three fast startup adjustments.
-
-After startup, the controller:
-
-- Samples the latest stable frame every five seconds.
-- Requires two consecutive bad samples before changing a setting.
-- Moves only one validated step during normal operation.
-- Waits ten seconds after a steady-state adjustment.
-- Defers changes while a triggered post-impact camera tail is being collected.
-- Recovers automatically when lighting improves.
+step at a time. It makes at most three fast startup adjustments before shot
+capture is armed. The selected manual exposure and gain are then locked for the
+life of that camera runtime, so every replay in the session uses the same image
+controls. There is no periodic exposure monitoring after startup.
 
 The last good setting is stored at:
 
@@ -260,7 +254,8 @@ stop the camera. Preview and raw frame capture continue, but camera-assisted
 horizontal launch, Club Path, and Attack Angle are withheld for that shot.
 Horizontal launch falls back to the IWR6843 path. The Camera tab reports
 **Lighting needed** and **Radar fallback active** so the operator can add or
-redirect light without restarting the session.
+redirect light. Restart OpenFlight after changing the lighting so startup
+calibration can select and lock a new setting.
 
 The standalone calibration sweep remains useful when diagnosing an unusual
 lighting installation:
--- a/docs/camera/README.md
+++ b/docs/camera/README.md
@@ -238,10 +238,16 @@ inside the configured frame period.
 At startup, the controller begins with the last good setting saved for the same
 resolution and frame rate. If that setting is not usable, it can jump several
 steps and recheck after `300 ms`, rather than waiting through one five-second
-step at a time. It makes at most three fast startup adjustments before shot
-capture is armed. The selected manual exposure and gain are then locked for the
-life of that camera runtime, so every replay in the session uses the same image
-controls. There is no periodic exposure monitoring after startup.
+step at a time. It makes at most three fast startup adjustments.
+
+After startup, the controller:
+
+- Samples the latest stable frame every five seconds.
+- Requires two consecutive bad samples before changing a setting.
+- Moves only one validated step during normal operation.
+- Waits ten seconds after a steady-state adjustment.
+- Defers changes while a triggered post-impact camera tail is being collected.
+- Recovers automatically when lighting improves.
 
 The last good setting is stored at:
 
@@ -254,8 +260,7 @@ stop the camera. Preview and raw frame capture continue, but camera-assisted
 horizontal launch, Club Path, and Attack Angle are withheld for that shot.
 Horizontal launch falls back to the IWR6843 path. The Camera tab reports
 **Lighting needed** and **Radar fallback active** so the operator can add or
-redirect light. Restart OpenFlight after changing the lighting so startup
-calibration can select and lock a new setting.
+redirect light without restarting the session.
 
 The standalone calibration sweep remains useful when diagnosing an unusual
 lighting installation:
--- a/src/openflight/camera/auto_exposure.py
+++ b/src/openflight/camera/auto_exposure.py
@@ -17,7 +17,6 @@ ExposureQualityStatus = Literal[
 ]
 ExposureRecommendation = Literal["brighter", "darker", "hold"]
 AutoExposureStatus = Literal[
-    "calibrating",
     "ready",
     "adjusting",
     "lighting_required",
@@ -173,7 +172,7 @@ def measure_exposure(image: np.ndarray) -> ExposureObservation:
     else:
         status = "marginal"
         recommendation = "darker" if clipped_pct > 2.0 or median > 180.0 else "brighter"
-        message = f"Exposure is usable; monitoring for a {recommendation} adjustment"
+        message = f"Exposure is usable; a {recommendation} setting may improve contrast"
 
     return ExposureObservation(
         sample_available=True,
@@ -199,36 +198,31 @@ def motion_blur_risk(exposure_us: int) -> Literal["low", "elevated", "high"]:
 
 
 class AutoExposurePolicy:
-    """Choose startup jumps and conservative steady-state adjustments."""
+    """Choose startup exposure steps and lock the first terminal decision."""
 
     def __init__(
         self,
         *,
         fps: float,
         startup_max_adjustments: int = 3,
-        steady_confirmations: int = 2,
     ):
         if startup_max_adjustments < 1:
             raise ValueError("startup_max_adjustments must be positive")
-        if steady_confirmations < 1:
-            raise ValueError("steady_confirmations must be positive")
         self.steps = exposure_steps_for_fps(fps)
         self.startup_max_adjustments = startup_max_adjustments
-        self.steady_confirmations = steady_confirmations
-        self._fast_reacquire_armed = False
-        self._reset_startup_state()
-
-    def _reset_startup_state(self) -> None:
-        """Enter startup convergence without rearming scene-change detection."""
         self._startup = True
         self._startup_adjustments = 0
-        self._pending_recommendation: ExposureRecommendation = "hold"
-        self._pending_count = 0
+        self._locked_decision: AutoExposureDecision | None = None
+
+    def _reset_startup_state(self) -> None:
+        """Enter startup convergence and clear any prior locked result."""
+        self._startup = True
+        self._startup_adjustments = 0
+        self._locked_decision: AutoExposureDecision | None = None
 
     def reset(self) -> None:
-        """Restart fast convergence after a material scene change."""
+        """Unlock the policy for a new camera startup."""
         self._reset_startup_state()
-        self._fast_reacquire_armed = False
 
     @property
     def startup(self) -> bool:
@@ -243,6 +237,9 @@ class AutoExposurePolicy:
         gain: float,
     ) -> AutoExposureDecision:
         """Return the next camera-control decision without applying it."""
+        if self._locked_decision is not None:
+            return self._locked_decision
+
         current_index = self._nearest_step_index(exposure_us, gain)
         current_step = self.steps[current_index]
         risk = motion_blur_risk(current_step.exposure_us)
@@ -259,36 +256,16 @@ class AutoExposurePolicy:
         if observation.acceptable:
             self._startup = False
             self._startup_adjustments = 0
-            self._pending_recommendation = "hold"
-            self._pending_count = 0
-            self._fast_reacquire_armed = True
-            return AutoExposureDecision(
+            self._locked_decision = AutoExposureDecision(
                 status="ready",
                 analysis_eligible=True,
                 message=observation.message,
                 observation=observation,
                 motion_blur_risk=risk,
             )
+            return self._locked_decision
 
-        if self._startup:
-            return self._startup_decision(observation, current_index)
-        if self._fast_reacquire_armed and self._is_material_change(
-            observation,
-            current_index,
-        ):
-            self._reset_startup_state()
-            self._fast_reacquire_armed = False
-            return self._startup_decision(observation, current_index)
-        return self._steady_decision(observation, current_index)
-
-    def _is_material_change(
-        self,
-        observation: ExposureObservation,
-        current_index: int,
-    ) -> bool:
-        """Return whether the measured scene needs a multi-step correction."""
-        target_index = self._startup_target_index(observation, current_index)
-        return abs(target_index - current_index) >= 2
+        return self._startup_decision(observation, current_index)
 
     def _startup_decision(
         self,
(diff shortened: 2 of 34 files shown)


johnpacino:
Great addition, a few blockers though: 

**1. Enrichment can replay “new shot” effects.**

[updateShot() increments shotVersion](https://github.com/jewbetcha/openflight/blob/ea43ef37934a930ebfa4742a9ccbfa454deb90fa/ui/src/stores/useShotStore.ts#L68-L86), which is also used to remount LivePanel and trigger Launch Daddy effects. If camera enrichment finishes within the 2.5-second isNewShot window, the flash/explosion runs twice. In-place enrichment should update the shot without advancing the new-shot signal; add a regression test.

 **2. The new shared worker needs concurrency/backpressure coverage.**

The enrichment queue is unbounded, while each item can wait seconds on hardware. The single added server test covers only one successful IWR shot. Add tests proving:
Two shots remain FIFO and use one worker.
Worker startup/enrichment failure still produces a usable OPS result.
Camera-only enrichment follows the same provisional/final contract.
Slow or stuck optional hardware cannot create an indefinitely growing Pi queue.

**3. Python formatting fails.**

ruff check passes, but uv run ruff format --check ... finds newly introduced formatting changes in:
src/openflight/iwr6843/doa.py
src/openflight/rolling_buffer/processor.py
src/openflight/server.py

**4. camera-only processing is invisible**

When camera enrichment runs without IWR, the processing overlay disappears after provisional OPS metrics and no pending indicator is displayed—even though camera work continues and may later alter the metrics.
No Blocking but documentation should be updated

5. The ~56 Hz standard FFT documentation remains stale in two places; 30,000 / 128 is approximately 234 Hz.

jewbetcha:
Addressed the blockers in 2924479 (merged with the two remote Copilot timeout commits in 75cb188):

1. shot_update now replaces metrics without advancing shotVersion; the regression test pins the new-shot signal.
2. Deferred enrichment now uses one FIFO worker with a two-shot waiting bound, non-blocking overflow fallback, and coverage for FIFO, startup failure, enrichment failure, camera-only flow, and stuck hardware.
3. The PR-touched Python files are Ruff-formatted.
4. Camera-only and combined IWR/camera work now show non-blocking OPS-ready progress.
5. The two standard FFT cadence references now say approximately 234 Hz.

Verification: 186 focused backend tests passed (1 hardware skip), all 244 UI tests passed, UI build/lint/Prettier passed, Ruff passed on modified files, and Pylint remains 9.49/10.

johnpacino (changes requested):
A new blocker was introduced with the latest checking

[P1 — blocking] The OPS-only failure path skips required shot finalization.
_emit_ops_enrichment_fallback() emits and forwards the provisional shot directly. On queue overflow, worker-start failure, or worker failure, this bypasses the normal launch-angle fallback, final carry/correction processing, and session_log.log_shot(). A valid OPS shot can therefore remain incomplete and be missing from the session JSONL.
Please separate optional hardware enrichment from required shot finalization, then ensure success, overflow, and worker-failure paths all finalize and persist the shot exactly once. Add regression assertions for session logging and normal OPS fallback metrics—not only the WebSocket event sequence and ball speed.

jewbetcha:
Addressed the P1 blocker in 2f813a4.

- Optional IWR/K-LD7/camera work is now isolated in _enrich_shot_from_optional_hardware().
- Required fallback angles, speed correction, spin/carry calculation, session_log.log_shot(), final WebSocket emit, and simulator forwarding now live in _finalize_shot_detected().
- Successful enrichment, queue overflow, worker-start failure, and enrichment failure each invoke required finalization exactly once. The old direct provisional emit fallback was removed.
- Regression tests assert one final event, one session-log entry, estimated launch-angle fallback, and positive finalized carry for every path.

Verification: 187 backend pipeline tests passed (1 hardware skip), Ruff and formatting passed, and Pylint remains 9.49/10.

johnpacino:
[P1 — blocking] Queue overflow breaks ordered finalization and can misassociate hardware artifacts.
When the enrichment queue is full, the newest shot is finalized immediately while older shots remain active/queued. Because IWR, K-LD7, and camera records derive shot_number from shots_detected + 1, the overflow finalization can increment that counter between an older shot’s hardware records and its final shot_detected entry. The overflow shot can also reach simulator connectors before the earlier shots.
Please give each shot a stable identity/sequence at detection and ensure final publishing and simulator forwarding preserve detection order, including overflow. Add a concurrent regression that blocks shot A, fills the queue, overflows with a newer shot, then releases the worker and verifies final WebSocket order, simulator order, and matching shot numbers/timestamps across IWR/camera/session records.

jewbetcha:
Addressed the ordered-finalization blocker in 6973608.

- Every shot now receives a stable session shot number at callback entry, before the provisional WebSocket emit or optional hardware work.
- IWR6843, K-LD7, camera, and final `shot_detected` records use that stable number; WebSocket payloads also carry the number and impact timestamp.
- Shot callbacks are serialized, and the bounded enrichment queue now applies backpressure when full instead of finalizing the newest shot out of order. OPS metrics are still emitted before that wait.
- Required final WebSocket publishing and simulator forwarding therefore remain FIFO through queue saturation and worker/startup failures.
- Added the requested concurrent regression: block A in IWR, queue B/C, submit D against the full queue, release A, then assert A→D order for final WebSocket updates and simulator forwarding plus matching shot numbers/timestamps across IWR, camera, and session records.

Verification: focused regression passes; full backend suite excluding unavailable local IWR binary fixtures is 1,589 passed / 6 skipped; UI is 244 passed plus successful build/lint/Prettier; Ruff passes; focused Pylint is 9.62/10.

johnpacino:
The concurrent regression now proves ordered final updates and resolves the previous artifact/simulator race. Two blockers remain:

[P1] Queue backpressure must not block the OPS callback. queue.put() can hold on_shot_detected() while optional hardware is slow, but the rolling-buffer monitor does not call trigger.reset() until that callback returns. This can prevent the production OPS path from rearming and lose subsequent shots. Keep the capture callback non-blocking and coordinate ordered final publication outside that thread.

[P2] Assign the stable shot identity before the raw OPS capture is logged. RollingBufferMonitor currently logs that artifact using len(self._shots) before the server assigns the new sequence. Clearing/deleting shots can therefore make OPS artifacts reuse numbers that no longer match IWR/camera/final records. Add the OPS capture to the cross-record identity regression.
The PR description should also be updated: it still says queue overflow uses an OPS-only fallback, while the current implementation applies blocking backpressure.

jewbetcha:
Addressed the remaining blockers and the follow-up camera/UI issues in e071d2b.

- **P1 / OPS rearm:** enrichment now uses `put_nowait()`. If the two-shot queue is saturated, the callback skips optional hardware for that shot and records its required finalization in a detection-order coordinator outside the capture callback. The concurrent regression blocks A in IWR, queues B/C, overflows D, proves D's callback returns while A is still blocked, then verifies final WebSocket and simulator order A through D.
- **P2 / cross-record identity:** `RollingBufferMonitor` assigns the monotonic shot number before appending or logging the raw OPS capture. Clearing/deleting visible shot history does not reset that identity. The regressions now match number/timestamp across raw OPS, IWR, camera, final session, WebSocket, and simulator records.
- The PR description now reflects non-blocking saturation rather than blocking backpressure.
- Ordered finalization continues after an individual finalizer failure, with explicit regression coverage.
- Live metric subtitles now wrap fully at 800x400, 800x480, and 1024x600; camera-source copy is compact and non-redundant.
- Camera exposure now calibrates synchronously before arming, locks the accepted exposure/gain for the runtime, removes steady-state adjustments, and refills the pre-trigger ring so calibration frames cannot leak into the first replay.

Validation: 1,592 backend tests passed / 6 skipped (excluding the unavailable real-capture fixture module); 244 UI unit tests and 26 Chromium E2E tests passed; build, ESLint, Prettier, Ruff, and Ruff formatting passed; full-project Pylint is 9.50/10 (focused changed modules 9.65/10).

johnpacino:
Thanks—the stable raw OPS identity, non-blocking callback, detection-order publication, and finalizer-failure handling now address the previous findings. One P1 blocker remains:

[P1] The ordered-finalization coordinator reintroduces unbounded memory growth.
_shot_finalization_order and _shot_finalization_ready are unbounded. If the earliest optional-hardware job never completes, every later saturated shot is registered and its full Shot object remains in _shot_finalization_ready, because the drain cannot advance past the missing head. I reproduced this with shot A held unready: the two-item enrichment queue remained bounded, but the coordinator accumulated 10,001 registrations and 10,000 full pending shots.
Please add a finite end-to-end enrichment deadline/watchdog. Once the head exceeds the deadline, finalize it OPS-only, ignore any late enrichment result, and continue draining. Add a regression where A never releases and many later shots overflow, asserting that callbacks return, coordinator state remains bounded, and valid OPS shots still finalize in order.

jewbetcha:
Addressed in `6fab8dd` with the watchdog rearm race follow-up in `3505496`.

- Ordered finalization now has a 20-second end-to-end enrichment deadline. If the head has not completed, it is finalized OPS-only and the coordinator continues draining.
- The coordinator itself has a hard four-shot capacity (active enrichment, two queued jobs, and one overflow result). Capacity pressure expires the oldest optional work OPS-only, so a rapid stream cannot accumulate full `Shot` objects behind a stuck head.
- Hardware enrichment mutates a dataclass copy. Results are copied back only while the original registration is still live; results arriving after timeout are ignored and cannot mutate, republish, or re-log the finalized shot.
- A single daemon watchdog is reused and safely rearmed after going idle.

The new regression holds shot A in enrichment past its deadline, submits 19 more shots while A remains blocked, and verifies:

- every OPS callback returns;
- the order, registration, and ready collections each remain at four entries or fewer;
- all 20 shots receive estimated fallback metrics and positive carry;
- session, WebSocket, and simulator finalization remains exactly-once and in detection order;
- releasing late hardware cannot alter the finalized shots.

Validation: full local backend suite is 1,592 passed / 6 skipped (excluding unavailable real-capture fixtures), focused Pylint is 9.71/10, and every PR CI job is green across Python 3.10-3.12, Pylint, UI build/lint/format, and Chromium E2E.

johnpacino:
Thanks—the camera exposure extraction is clean, and the deadline/cap now prevents the stuck optional-hardware scenario from growing the coordinator indefinitely. One P1 remains:

[P1] Capacity eviction runs required finalization synchronously on the OPS callback thread.
When _register_shot_for_finalization() takes the coordinator above capacity, it calls _drain_ordered_shot_finalizations() directly. Registration occurs inside on_shot_detected(), so the fifth callback can perform fallback calculation, session logging, WebSocket emission, and synchronous simulator forwarding before returning. I reproduced this with a blocked finalizer: the fifth OPS callback did not return while finalization was blocked. This can again delay trigger.reset() and OPS rearming.

Please make the watchdog/dedicated coordinator thread the exclusive owner of draining; callback paths should only register/queue and notify it. Add a capacity-pressure regression that blocks _finalize_shot_detected() and proves the fifth on_shot_detected() call still returns immediately.

[P2] Overflow still advertises hardware processing that was skipped.
The initial event sets IWR/camera pending flags before queue admission. If put_nowait() fails, that shot skips optional hardware but the kiosk continues displaying “Receiving IWR radar dump / processing camera capture” until ordered finalization reaches it. Please base pending state on successful admission or immediately publish an explicit OPS-only/skipped state.
The exposure redesign is now cleanly removed. The remaining metric-subtitle wrapping/copy change is still independent of the latency work and should be split into its own UI PR.

jewbetcha:
Addressed the latest review in `80cbd93` (with the CI-only Prettier follow-up in `5d51458`).

- **P1 / callback isolation:** a dedicated `shot-finalization` thread is now the exclusive ordered drainer. OPS callback paths only register or queue work and notify the coordinator; fallback calculation, persistence, WebSocket publication, and simulator forwarding never run on the capture callback. The new capacity-pressure regression blocks `_finalize_shot_detected()` and proves the fifth `on_shot_detected()` call still returns while finalization remains blocked.
- **P2 / skipped hardware feedback:** queue saturation and worker-admission failure now emit an immediate timestamp-matched `shot_update` with `pending: {}` and explicit `enrichment.status: skipped` metadata before required ordered finalization. The UI regression proves this clears the matching IWR/camera progress state without adding or replaying the shot.
- **Subtitle split:** all five live-metric subtitle files were removed from this PR in `d2e5834` and moved unchanged into standalone PR #246.
- The PR description now reflects the exclusive coordinator, explicit skipped state, and both split PRs.

Validation: 1,593 backend tests passed / 6 skipped locally after excluding the four unavailable real-capture assertions whose `.l3dump` fixtures are absent; all 245 UI unit tests passed; focused Pylint is 9.70/10. PR #240 and subtitle PR #246 are both green across Python 3.10-3.12, Pylint, build, lint, format, and Chromium E2E.

johnpacino (approved):
Thanks for the rounds of feedback. Looks good!

PR #243, opened 2026-08-27
feat(kiosk): replace players with persistent server-owned profiles

Cormac131:
## What does this PR do?

Replaces the kiosk's browser-local "player" roster with a server-owned **profile** system for shot attribution.

A profile is a named context — a person *or* a place — with a stable id. The server is now the single source of truth:

- Roster and active profile live in `~/.config/openflight/profiles.json` (atomic writes, seeded default if missing/corrupt)
- One authoritative `profiles` snapshot on connect and after every mutation
- Mutations: `set_active_profile` / `add_profile` / `rename_profile` / `remove_profile`
- Shots and swing-speed events stamp `profile_id` + `profile_name`; filtering is by exact id, not name
- The Profiles tab lists the roster with shot counts, rename, and remove (active/last profile cannot be deleted)
- Add/rename uses a full-screen on-screen keyboard (`inputMode="none"`) so Chromium `--kiosk` on the Pi is usable without a system keyboard

**Breaking:** `set_player` / `player_changed` are gone. `Shot.player_name` is replaced by `profile_id` + `profile_name`. Existing browser-local player rosters in `localStorage` are discarded (clean break, no migration). GSPro/`sim` `Player` fields are unchanged — they are an external wire protocol.

image
image

## Why was this required?

A player was a bare name string in the browser. That broke three things on a shared kiosk:

1. **A name is not an identity.** Renaming orphaned every shot recorded under the old name.
2. **The roster was browser-local.** A reflashed Pi, a second browser, or any future server-side setting had nothing to attach to.
3. **Two sources of truth raced.** `session_state.player_name` and `player_changed` could disagree, which is why `playerSocketSync` existed.

Without this, session stats stay fragile, rename is unsafe, and later per-profile settings have nowhere to live. "Player" was also the wrong word for a range bay or home net.

## Automated tests

**Python**
- `tests/test_profiles.py` — seed/corrupt fallback, add/rename/remove/set-active, duplicate names, 40-char cap, roster cap of 12, reject active/last/unknown, settings round-trip, snapshot shape, atomic write (failed `os.replace` leaves the previous file intact)
- `tests/test_server.py` — `TestProfileStamping` (shot/swing-speed emit profile fields), `TestProfileSocketHandlers` (snapshot on get/add/rename/remove/set-active, rejected mutations rebroadcast unchanged state), shot/swing-speed stamped with the active profile, `clear_session` scoped to one profile id, case-differing names do not collide

**UI**
- `ProfilesPanel` — shot counts, active pressed state, remove hidden on active/last, rename always available, skeleton until snapshot
- `ProfileNameDialog` — add vs rename copy, 40-char cap, blank confirm disabled, on-screen keyboard present with `inputMode="none"`
- `useProfileStore` — snapshot apply, ignore malformed payloads
- Live / Stats / Shots panels filter by `profile_id`; i18n shot-count strings
- Mock server handlers/session updated to the new socket contract

**E2E** (`ui/tests/e2e/app.spec.ts`)
- Profiles tab and Add profile control
- Select profile → Live, no delete on the active profile
- Clear session confirms and only removes that profile's shots
- Rename via dialog and Enter
- Type a name with the on-screen keyboard at 800×400
- Rename keeps the profile's shots

## Manual (human) testing

Ran the kiosk UI against the mock server (`cd ui && npm run dev`) and walked the Profiles flows by hand, including the 800×400 viewport used on smaller kiosks.

- Opened **Profiles**: seeded **Profile 1**, Add in the header, skeleton gone after the first `profiles` snapshot
- Added **Alex** with the on-screen keyboard (QWERTY → auto-unshift after first letter, confirm stays in view). Native OSK did not appear (`inputMode="none"`)
- New profile became active; Live subtitle showed **Alex**; simulating a shot attributed it to Alex only
- Switched back to Profile 1 from the roster (opens Live); remove was hidden on the active card and shown on the other
- Renamed a typo (**Rnage** → **Range**); shots stayed on that profile and the Shots list showed the new name
- **Clear session** copy named the active profile, Cancel left Stats intact, Confirm wiped only that profile and returned to Live Ready; the other profile's shots were still there
- Could not remove the last remaining profile
- Checked EN/ES/FR/PT strings for Profiles / Add / Rename / Clear session (no leftover "player" copy in those flows)

Not exercised on a Pi with real radar in this pass — keyboard and roster behavior were verified in mock/dev. The Pi kiosk path is the reason the OSK exists (Chromium `--kiosk` has no system keyboard).

## Checklist


johnpacino (changes requested):
Thanks for the work here—the server-owned profile direction is sound and the test coverage is substantial. I’m requesting changes for the following items before approval:

**1. [P1] Isolate profile persistence during E2E tests**

ProfileStore defaults to ~/.config/openflight/profiles.json, while the E2E reset helper clears profiles and removes all but the default. Running the E2E suite can therefore create, modify, or delete the developer’s real profile roster.
Please make the profile path configurable and start every E2E backend with a unique temporary path per test worker. Add coverage proving test execution cannot touch the default user configuration.

**2. [P1] Do not allow profile deletion to orphan session shots**

The UI permits removing an inactive profile even when it has shots. ProfileStore.remove() deletes only the roster entry, leaving those shots associated with an ID that is no longer selectable. Re-adding the same name generates a new ID and does not reconnect them.
Please either reject removal while the profile has session rows—which is my recommendation—or implement an explicit, atomic removal flow that also handles those rows. Cover the chosen behavior in backend and browser tests.

**3. [P2] Restore the clear-session modal styling and interaction coverage**

ClearSessionDialog.tsx now uses clear-session-modal* class names, but panel.css does not define them. At the 800×480 kiosk viewport, the confirmation renders as inline content without the intended modal positioning or scrim.
Please restore the appropriate styles and add a browser test that verifies the overlay/dialog layout and cancellation behavior at 800×480. Text-presence assertions alone do not catch this regression.

**4. Rebase onto current main**

This branch is currently 54 commits behind and conflicts in:
docs/CHANGELOG.md
src/openflight/server.py
ui/src/services/socketService.ts
During resolution, preserve main’s two-phase shot delivery: stamp the active profile before the initial OPS shot emit, retain the final shot_update event after optional hardware enrichment, and keep the UI’s handleShotUpdate path. Please rerun all required checks against the rebased head.

**5. Keep repository-wide formatting policy out of this feature PR**

Please remove .editorconfig, .gitattributes, and the endOfLine change in ui/.prettierrc. These establish repository-wide editor and line-ending policy and should be submitted in a dedicated maintenance PR. The profile-related README and changelog updates are in scope and should remain.

**Non-blocking follow-up**

**6. [P3] Make mutation and persistence one transaction**

ProfileStore releases its lock between mutating in-memory state and saving it. Concurrent profile commands can therefore persist an older snapshot after a newer one, causing the roster to revert after restart. This is unlikely on a single kiosk, so I’m not holding the merge for it, but it should be tracked and resolved by keeping mutation, snapshot creation, and persistence ordering atomic.

I also reviewed the PR’s size: 66 files and 3,484 changed lines. Although larger than ideal, most of that volume is justified because the existing “player” concept was spread across the backend, socket contract, UI, mocks, translations, tests, and documentation. I’m not requesting that the profile migration be split solely because of its size; the unrelated repository-wide formatting configuration is the scope exception noted above.

Once the blocking items are addressed and CI is green on the rebased head, please request another review.

johnpacino (changes requested):
Thanks—the deletion safeguards, clear-session modal, main integration, and formatting-policy cleanup all look good. I have one remaining issue from the original E2E-isolation blocker:

**[P1] Prevent E2E from reusing an existing backend**
ui/playwright.config.ts still sets reuseExistingServer: !process.env.CI for the backend. When E2E is run locally while OpenFlight is already running on port 8080, Playwright reuses that process instead of starting the backend with OPENFLIGHT_PROFILES_PATH. resetSession() can then clear the real session and remove every non-default profile from the real persisted roster.
This is not a new regression from the latest commit; it is a remaining gap in the isolation fix. Please either set reuseExistingServer: false for the backend entry—which will fail safely if port 8080 is occupied—or use a dedicated E2E backend port. Please also update the isolation test to cover this condition.

Once this is updated, we should be good to go!

johnpacino (changes requested):
Thanks—the backend isolation fix is correct and resolves the remaining P1. One small CI issue remains:

ui/tests/e2e/isolateProfilesPath.test.ts is a Vitest test inside Playwright’s tests/e2e directory. Playwright also discovers .test.ts files, tries to execute the Vitest describe(), and fails before collecting any browser tests. This reproduces locally and is why the latest E2E check is red.

Please either move the Vitest test outside tests/e2e, or configure Playwright with testMatch: '**/*.spec.ts'. Once the E2E check is green, I’m good to approve.

johnpacino (approved):
Looks good, thanks!

jewbetcha (approved):

PR #244, opened 2026-08-27
fix: lock camera exposure after startup

jewbetcha:
## What does this PR do?

- Runs automatic exposure calibration synchronously once during camera startup.
- Locks the first terminal exposure decision and its manual exposure/gain controls for the lifetime of the camera runtime.
- Removes the periodic five-second exposure monitor and steady-state adjustment path.
- Discards calibration frames and refills the pre-trigger buffer at the locked settings before GPIO shot capture is armed.
- Updates the Camera tab, CLI descriptions, and camera documentation to describe the startup-only behavior.

## Why was this required?

Exposure could change after startup as ambient conditions varied. That made otherwise similar camera replays alternate between analysis-eligible and rejected, and a replay could contain pre-trigger frames captured across different calibration settings. A single startup calibration makes replay quality deterministic for the session. Operators can restart OpenFlight after a material lighting change to select a new locked setting.

## Behavior and tradeoff

Startup may take up to three fast exposure steps plus one pre-trigger-buffer refill before the trigger is armed. In exchange, no shot is accepted until calibration is complete and every replay in that runtime uses one fixed exposure/gain pair.

## Automated tests

- Proves an accepted startup setting remains locked when later frames become dark.
- Proves a lighting-required result remains locked until the next runtime reset.
- Proves startup waits for calibration instead of launching a background monitor.
- Proves calibration frames are discarded before the locked pre-trigger buffer is filled.
- Isolated camera exposure/runtime suite: 52 passed.
- Ruff check and format verification pass on all changed Python files.

## Manual (human) testing

- Real camera hardware was not available locally; final validation on the OV9281/Pi capture stack remains pending.

## Checklist

- [x] Single feature/fix
- [x] Automated tests included
- [x] Documentation updated
- [x] No new dependencies
- [x] No unrelated performance-pipeline changes

### DIFF
--- a/docs/camera/README.md
+++ b/docs/camera/README.md
@@ -238,16 +238,10 @@ inside the configured frame period.
 At startup, the controller begins with the last good setting saved for the same
 resolution and frame rate. If that setting is not usable, it can jump several
 steps and recheck after `300 ms`, rather than waiting through one five-second
-step at a time. It makes at most three fast startup adjustments.
-
-After startup, the controller:
-
-- Samples the latest stable frame every five seconds.
-- Requires two consecutive bad samples before changing a setting.
-- Moves only one validated step during normal operation.
-- Waits ten seconds after a steady-state adjustment.
-- Defers changes while a triggered post-impact camera tail is being collected.
-- Recovers automatically when lighting improves.
+step at a time. It makes at most three fast startup adjustments before shot
+capture is armed. The selected manual exposure and gain are then locked for the
+life of that camera runtime, so every replay in the session uses the same image
+controls. There is no periodic exposure monitoring after startup.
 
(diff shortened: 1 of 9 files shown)


johnpacino (approved):
Reviewed exact head 83db1e6. Startup calibration is synchronous, the terminal exposure decision remains locked, calibration frames are discarded before the capture buffer is refilled, and optional camera measurements continue to fail safely without losing valid OPS shots. Exact-head CI is green, and the functionally identical camera code successfully calibrated and armed the OV9281 on the target Pi.  

The remaining exception-handling concern is a non-blocking P3 resilience case: an unexpected calibration exception is logged while capture continues with camera analysis normally ineligible. That can be hardened separately. Approved.

PR #249, opened 2026-08-28
fix(camera): project attack angle onto full horizontal speed

HarjotDhanota:
## What does this PR do?

Fixes the attack-angle projection in the camera club delivery. Vertical speed was divided by the forward component alone rather than the full horizontal speed, which inflates attack angle by 1/cos(club path) — zero on a dead-straight path, growing with path angle. Measured on a 21-shot session: +0.05° mean, 0.35° max at the observed paths, and about 0.6° at a 30° path.

## Why was this required?

The error is small but systematic and correlated with club path, so it biases exactly the shots where path matters most. Two shots with identical true attack angles read differently if their paths differ.

## Automated tests

`tests/test_camera_club_delivery.py` gains three cases: the corrected projection against hand-computed 3D velocities, agreement with the old form at zero path, and that the divergence between old and new forms matches 1/cos(path) exactly.

## Manual (human) testing

The mean and maximum deltas come from replaying the 21-shot reference session through both forms and differencing per shot; the deltas match the predicted 1/cos(path) inflation. Club path is computed from the same inputs and is untouched by this change.

## Checklist

- [x] **Single feature/fix** — one projection fix
- [x] **Automated tests included** — three regression cases
- [x] **Manual testing described** — session replay above
- [x] UI builds — n/a, no UI changes
- [x] UI lint passes — n/a, no UI changes
- [x] Updated docs and CHANGELOG


jewbetcha (approved):
Nice

PR #266, opened 2026-09-14
fix(ballistics): replace Hill lift form with Ferguson quadratic Cd/Cl

### LINKED ISSUE #230: Cl Hill form is monotonic, so it cannot represent the measured lift peak near S~0.5
### Description

Split out of #218 / #229 so it doesn't expand that PR.

`ballistics.py` models lift as a saturating Hill function:

```python
CL_SATURATION = 0.32
CL_HALF_SP = 0.15
# Cl = CL_SATURATION * Sp / (CL_HALF_SP + Sp)
```

That form is **monotonic increasing and asymptotically flat**. Measured golf-ball `Cl` is not: it rises, **peaks around spin ratio S ≈ 0.5, then falls**. So above roughly S ≈ 0.35 no choice of the two constants can match reality — the shape is wrong, not the numbers.

This is measurable on the capture already in the repo. Same 24 shots, same RK4 integrator, same `resolve_launch`/`simulate` path — the only thing changed is the `Cd`/`Cl` functional form, each fitted to its own optimum:

| form | OVERALL | driver | 7-iron | pitching wedge |
|---|---|---|---|---|
| current: `Cd = a+bS`, `Cl = c·S/(d+S)` (4 param) | 3.97 | 1.75 | 2.96 | **6.26** |
| quadratic: `Cd = a+bS+cS²`, `Cl = d+eS+fS²` (6 param) | **1.99** | 1.07 | **0.98** | **3.33** |

Roughly half the overall error, and it nearly halves the wedge column — which is exactly where #229 stays weak. Wedges in that capture sit at `Sp` 0.376-0.632, i.e. right where the Hill form flattens but real `Cl` is peaking and turning over.

Two independent reasons to think the quadratic is the right shape rather than just more parameters:

1. **Published models use it.** Ferguson, McNally & McPhee (2022, ISEA 14, [open access](https://docs.lib.purdue.edu/resec-isea/2022/session05/4/)) fit exactly `Cd = a+bS+cS²`, `Cl = d+eS+fS²` to **N = 1040 shots** (GCQuad launch conditions + FlightScope X3 carry/apex/offline, Titleist Pro V1, S = 0.02-0.75) and report 2.51 m / 2.74 yd carry MAE. Quintavalla (2002) and Smits & Smith (1994) are the other standard references, both non-monotonic in S. Lyu/Kensrud/Smith (2018, ISEA) also state outright that lift "had a non-linear dependence on spin (fit with a 2nd order polynomial)".

2. **Fitting it here reproduces their constants.** Fitted to our 24 shots the quadratic gives `[0.1346, 0.9438, -0.7351, 0.0548, 1.2258, -0.9196]`, against their published `[0.1304, 0.9287, -0.8259, 0.0504, 1.2031, -1.1490]` from 1040 shots. Two independent datasets converging on nearly the same coefficients is decent evidence the form is right and the fit isn't noise.

### What did you expect to happen?

Not necessarily a change — mainly a decision, since it's architectural and I'd rather ask than assume.

**Option A — switch `Cl`/`Cd` to 2nd-order polynomials in S.** Better accuracy, aligns with the literature the file already cites, keeps everything else (RK4, Magnus geometry, spin decay) untouched. Costs: 4 constants become 6, and `Cl` must be clamped at `>= 0` since a downward parabola eventually goes negative outside the fitted range.

**Option B — leave it.** Defensible if you'd rather not widen the parameter count on 24 shots of ground truth, and the current form is fine for driver/iron `Sp`.

Related structural gaps I noticed while looking, listed only so they're on the record — not proposing to fix them here:

- **No Reynolds-number dependence anywhere** (`grep -ci reynolds ballistics.py` = 0). Ferguson, Quintavalla and Smits & Smith all include it; Lyu et al. measured `Cd` falling 0.5 -> 0.2 across the drag crisis. `Cd = f(S)` alone conflates speed and spin.
- **No moment coefficient.** Spin decay is a fixed 4%/s exponential rather than aerodynamic; Ferguson fits `C_M = g·S`.
- **Spin axis is constant through flight** — no precession. Fine for carry, matters for curve.

Happy to implement whichever you prefer, or to leave it as a documented limitation. Worth noting these need refitting together if #169 (altitude air density) lands, since air density and the coefficients trade off against each other.

### Area

Ballistics / physics model

cvrt-jh:
## What does this PR do?

Replaces the aerodynamic coefficient forms in `ballistics.py`. Drag was linear in the spin parameter and lift a Hill-type saturating curve, four fitted constants between them. Both are now second-order polynomials in the spin parameter, using the coefficients Ferguson, McNally & McPhee (2022, ISEA 14, open access) published from their 1040-shot fit. Nothing was re-fitted here; the six numbers are theirs as printed.

Fixes #230, Option A, the variant we recommended in the validation comment there.

Five files, Python and docs only:

- `src/openflight/ballistics.py`: two coefficient tuples, a clamp constant, and the two evaluators. Integrator, Magnus geometry, spin decay and ball constants untouched.
- `tests/test_ballistics_trackman_regression.py`: budgets ratcheted down to the new baseline.
- `scripts/analysis/sweep_ballistic_coeffs.py`: the fitting tool referenced the removed constant names and would have crashed, so it now sweeps the six polynomial coefficients.
- `docs/how-it-works/ballistics.md` and `docs/reference/constants.md`: model equations and constants tables (both still showed pre-#229 values).

## Why was this required?

The Hill form is monotonic and flattens by construction. Measured lift rises, peaks near a spin parameter of about 0.5, then falls. No choice of the two Hill constants produces that shape, so #229 could only trade carry accuracy against drag, leaving every club about 4 yd low at apex.

Both arms below run the unchanged `simulate()` path fed TrackMan's measured launch conditions, so the only difference is the body of `_cd` and `_cl`.

**Committed 24-shot capture** (the #219 regression fixture):

| club | Hill (#229) | quadratic |
|---|---|---|
| 7-iron | 2.96 | 1.33 |
| driver | 1.75 | 2.38 |
| pitching wedge | 6.26 | 4.48 |
| overall carry RMSE | 3.97 | 2.90 |

Driver gets slightly worse on this fixture; its budget was held, not raised, and still passes.

**Independent 5-session archive** from @johnpacino, 279 clean shots these coefficients never saw, with per-shot spin axis and per-session air density:

| | Hill (#229) | quadratic |
|---|---|---|
| carry RMSE | 3.48 | 1.98 |
| carry MAE | 2.79 | 1.61 |
| apex RMSE | 4.41 | 0.97 |
| apex bias | -3.99 | -0.59 |

Per club carry RMSE, Hill vs quadratic: 9i 3.22 / 1.61 (n=176), driver 4.10 / 1.90 (33), 5i 3.35 / 2.58 (32), 7i 3.90 / 1.96 (30), sand wedge 4.73 / 4.92 (8). On all 416 shots including mishits: 4.83 / 2.25.

Apex is the strongest evidence: nothing here was ever fitted against it, and the bias column shows the Hill form was structurally low, not noisy.

**One implementation detail worth a look.** The spin parameter is evaluated every integration step, and for wedges near apex it climbs past the fitted range (0.02 to 0.75) because speed falls faster than spin decays. Extrapolated, the parabolas send lift negative above 1.09 and drag negative above 1.25, both reachable on a high lob. So above `SP_FIT_MAX = 0.75` both curves are held at their end value. That clamp is why the numbers above differ slightly from the #230 comment, which ran the raw polynomials (carry 1.91 to 1.98, apex 1.00 to 0.97, sand wedge 3.82 to 4.92 on 8 shots). Lift is also clamped at zero, and zero spin still gives zero lift.

## Automated tests

No new test file. The #219 guard covers exactly this surface; its budgets were ratcheted: overall 5.0 to 3.5, 7-iron 4.0 to 2.0, pitching wedge 7.5 to 5.5, driver held at 3.0. The before/after table is in the module comment.

Verified they can still fail: lowering the linear lift coefficient from 1.2031 to 1.0031 fails all four assertions (overall 10.94 yd against 3.5), and restoring it goes green.

Full suite locally on Python 3.11: 1691 passed, 10 skipped, zero failures.

## Manual (human) testing

- Re-ran both validations against the patched module rather than a monkeypatch, the committed capture through `validate_ballistics.py` and the archive through our own harness, reproducing every number above.
- Controls: a zero-lift patch moves capture RMSE to 76.55 yd and restores exactly; a Hill re-patch reproduces 3.97 exactly; the clamp returns identical lift at 0.75 and 1.5, zero at zero spin, and zero for a deliberately overshooting negative term.
- Smoke-ran the adapted sweep script on the capture with a tiny search budget: runs end to end, prints six coefficients, fits to 2.03 yd near the published values, consistent with the 24-shot fit in #230.
- Docs build clean with the strict flag. Pylint 9.52, identical to main. Ruff check has one pre-existing import-order finding in the sweep script's untouched import block, and ruff format flags pre-existing lines in both Python files; no line this PR adds is flagged. Left alone rather than mixing a formatting sweep into a physics change.

Caveat: Python 3.11 only locally; 3.10 and 3.12 rely on CI. Pure float arithmetic, no version sensitivity expected.

AI assistance disclosure, per AI-POLICY.md: worked with Claude throughout, on the code, the validation runs and this text. The physics decisions, the clamp, and the numbers reported here are ours.

## Checklist

- [x] **Single feature/fix**: the Cd/Cl functional form, plus the test, tool and doc updates the change forces
- [x] **Automated tests included**: ratcheted regression guard, proven able to fail
- [x] **Manual testing described**
- [x] Python tests pass: 1691 passed, 10 skipped
- [x] Pylint passes: 9.52/10, unchanged from main
- [x] Ruff passes on all lines this PR adds (one pre-existing finding in an untouched import block noted above)
- [ ] UI builds: n/a, no UI changes
- [ ] UI lint passes: n/a, no UI changes
- [x] Updated docs: both constants tables and the model description


johnpacino (changes requested):
The model change looks good: I confirmed the committed 24-shot carry results, and the cap addresses the negative-drag case we discussed.

One request before I approve: please add automated tests for the new safeguards and the apex improvement. The current carry tests still pass if the cap is removed, even though drag becomes negative at a spin ratio of 1.5. A small test covering zero-spin lift, both curves above the 0.75 cap, and positive drag for a low-speed lob-wedge shot would protect that behavior.

The committed TrackMan CSV also has apex measurements (Max Height - Height, in feet). Please add an apex error budget using that fixture; I independently confirmed a substantial improvement for the 7-iron and pitching wedge. There’s no need to add the full five-session archive to this PR.

With those tests and green CI on the update, I’d be comfortable approving.

cvrt-jh:
Both added, and the clamp gap was real. With `SP_FIT_MAX` removed from `_cd` and `_cl` the entire suite still passed, so the guard had no coverage at all.

**Safeguard tests** (`TestAeroCoefficientSafeguards`, 6 tests): zero-spin lift, non-negative lift and positive drag above the fitted range, both curves held flat past `SP_FIT_MAX`, and the fitted region below it unchanged. Reverting the clamp fails three of them.

Your Sp 1.5 figure checks out: raw `Cd` is -0.3503 there, crossing zero at Sp ~1.251, and raw `Cl` goes negative above Sp ~1.09.

One thing worth flagging on the lob-wedge case, because it did not work the way I expected. I wrote that test, and it passed with the clamp removed. Removing the clamp *lowers* peak Sp rather than raising it (1.510 to 1.226 for 55 mph / 40 deg / 10000 rpm), since less drag keeps the ball faster, and the unclamped flight then never reaches negative drag at all: 0 negative-`Cd` evaluations out of 6076 integration steps. So a flight-level assertion cannot act as a red control here. I kept the shot as a reachability test, asserting that a realistic lob wedge does push Sp past `SP_FIT_MAX` at all, with the limitation written into the docstring. The coefficient-level tests are what actually hold the guard.

I also discarded an assertion of my own along the way: "speed never increases in flight" is wrong, since a descending ball gains speed under gravity either way (31 rises clamped, 30 unclamped). It would have failed on correct code.

**Apex budget**, from the committed capture's `Max Height - Height`:

| club | Hill | quadratic | budget |
|---|---|---|---|
| 7-iron | 18.15 | 1.53 | 2.0 |
| pitching wedge | 18.00 | 2.81 | 3.5 |
| driver | 5.08 | 3.17 | 4.0 |
| overall | 15.05 | 2.56 | 3.0 |

All feet, n=24. Restoring the Hill form fails all four budgets. This matches the substantial 7-iron and pitching wedge improvement you saw independently.

Apex is plumbed through `load_trackman`/`validate_tm_inputs` rather than parsed separately, so the test and the manual validator cannot disagree about the fixture, per that module's existing note. The fields are `Optional` and default to `None`, so the paired comparison CSV path is unaffected and the per-shot CSV columns are unchanged. There is also a coverage test pinning apex presence on all 24 shots, so a fixture that loses the column fails loudly instead of turning the budgets into vacuous passes over an empty list.

Budgets are seeded above measured to absorb float and platform drift, matching the carry budgets; ratchet down, never up.

No five-session archive here, as you asked.

**Verification:** 1565 passed, 7 skipped on py3.11. That is below the 1691 I reported on the last round: clean `main` is now 1554/7 after the recent commits, and this branch adds 11. Rebased onto current `main`, with the pre-existing commit's diff unchanged apart from blob hashes and no competing commits on any touched file. Docs build clean under `--strict`. The `ruff`/`pylint` findings on `validate_ballistics.py` (unused `json` and `SPIN_CONFIDENCE_HIGH` imports, plus one formatting complaint) are present on clean `main` and left alone rather than mixed into this change.

Worked with Claude throughout, on the code, the validation runs and this text. The physics decisions, the clamp, and the numbers reported here are ours.

johnpacino (approved):
Thanks for updating, this looks good!

johnpacino:
@jewbetcha - can I get your Eyes on this one?

jewbetcha (approved):
LGTM

